PR #52, opened 2020-09-11
Update release script to transpile esmodule code into commonjs for node

ChrisDufourMB:
## Fixes

_Please link to the issue this PR addresses_

## Description

WIP - This is disgusting but it works

How to generate a usable index.js:
- run `npm install @babel/core @babel/cli @babel/preset-env @babel/plugin-transform-modules-commonjs` in the project root
- Run `deno bundle -c tsconfig.json mod.ts | node_modules/.bin/babel --presets=@babel/preset-env --plugins=@babel/plugin-transform-modules-commonjs --no-babelrc -- > index.js`
This will bundle the deno ts code into a js ESModule, then use babel to convert the ESNext features using polyfills, and convert the ESModule into a commonjs module, which node can consume
- Create a new directory, then `cd` into it
- Copy `index.js` into this new directory
- Create the following typescript file:
`serialize-test.ts`
```ts
import { strict as assert } from 'assert';
// @ts-ignore
import { Serializable, SerializeProperty } from "./index";

class Test extends Serializable {
    @SerializeProperty()
    testName = "toJson";
  }
  // @ts-ignore
  assert.strictEqual(new Test().toJson(), `{"testName":"toJson"}`);
  // @ts-ignore
  console.log(new Test().toJson())
  // @ts-ignore
  const test = new Test().fromJson(`{"testName":"fromJson"}`);
  assert.strictEqual(test.testName, "fromJson");
  console.log(test)
```
- And the following ts config file:
`tsconfig.json`
```JSON
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "experimentalDecorators": true,
    "forceConsistentCasingInFileNames": true
  }
}
```
- install typescript `npm install -g typescript`
- run `tsc` to compile `serialize-test.ts` into `serialize-test.js`
- run `node serialize-test.js`, you should see the following output:
```js
{"testName":"toJson"}
Test { testName: 'fromJson' }
```
## Proposed changes in this PR

- foo
- bar
- baz

## Things to look at

- [ ] Test coverage
- [ ] Code Style
- [ ] Documentation (READMEs etc)

PR #53, opened 2020-09-11
Node deploys

hardy925:
## Description

Adding automated NPM deploys

To test this, create a new project: 
```bash
$ mkdir test
$ cd test
$ npm init
$ npm i @gamebridgeai/ts_serialize typescript
```
add a `tsconfig.json` file with
```json
{
    "compilerOptions": {
        "experimentalDecorators": true,
    }
}
```
add the test code to the `test.ts` file
```ts
import { Serializable, SerializeProperty } from "@gamebridgeai/ts_serialize";
  @SerializeProperty()
  testName = "toJson";
}
console.assert(new Test().toJson() === `{"testName":"toJson"}`);
console.log(new Test().toJson());
const test = new Test().fromJson(`{"testName":"fromJson"}`);
console.assert(test.testName === "fromJson");
console.log(test);
```
and compile the file

```
$ ./node_modules/.bin/tsc -p tsconfig.json
```
then run the compiled file with node

```
$ node test.js
```
Screen Shot 2020-09-14 at 2 15 05 PM

- automated NPM releases

## Things to look at

- [x] Test coverage
- [x] Code Style


types.d.ts
```diff
@@ -0,0 +1,53 @@
+export abstract class Serializable {
```
ChrisDufourMB: This needs to be renamed to `index.d.ts` and wrapped in `declare module "ts_serialize" {` so that typescript will automatically load the types for the `dist` module, and when a user runs `import { Serializable, SerializeProperty } from "ts_serialize";` it will be associated with these definitions

.github/workflows/scripts/create_npm_package_file.sh
```diff
@@ -15,6 +15,7 @@ cat <<EOF >> "package.json"
   "version": "$1",
   "description": "A zero dependency library for serializing data.",
   "main": "index.js",
+  "types": "types.d.ts",
```
ChrisDufourMB: Because our `main` is `index.js` if we rename `types.d.ts` to `index.d.ts` then we don't even have to specify a types file:

> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

ChrisDufourMB (commented):
Everything here looks good!

Tested the babel compilation locally with the test files from #52 and it works perfectly.

.github/workflows/scripts/babel_ts_serialize.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const [, source] = await Deno.bundle("./mod.ts");
+const { code } = babelCore.transformSync(source, {
+  filename: "dist/index.js",
+  presets: [presetEnv],
+  babelrc: false,
+  configFile: false,
+});
+
+console.log(code);
```
ChrisDufourMB: Just a quick thing to note: This will add a newline to the end of the output produced by babel. If we want to avoid that it's possible to write directly to stdout:
`await Deno.stdout.write(new TextEncoder().encode("Some Text"));`

hardy925:
@ChrisDufourMB 

I am not confident enough to merge this PR, right now it would include the `types.d.ts` that I think should be a generated CI file when we build the nodeJS dist. The CI action to deploy to NPM is not there yet either

I would say this brings us like 30-40% to completion. And acts as a PoC that _it can work_, but this still needs a lot of work.

What are your thoughts?

ChrisDufourMB:
This makes sense to me. If the goal of this PR is to have a completed node module that can be consumed then we still need the features that you mentioned: generated types and the NPM action.

As it stands now we have a manually created types file that we can't be sure is correctly formatted (we have `export type`, but maybe it should be `export declare type` as it is in the source code? https://github.com/GameBridgeAI/ts_serialize/blob/develop/serialize_property.ts#L16). We may want to do some experiments with `tsc -d` to see what it produces given certain exports.

The way we are currently deploying the package (a tar.gz file) is also not correct and needs to be refactored into a proper npm deployment to a repository somewhere.

I think we will also want to write an implementation test to verify that the produced node module can be consumed by node. Maybe it's possible to write a test that both Deno and nodejs+ts can run to confirm that it works exactly the same in both?

hardy925:
The `declare` in the source is an error on my part. `declare` is used for telling the compiler the type will be available at runtime in the global scope. So if a `js` package is included in `ts` without types and you know a function will exist in the global scope from the `js` package you can `declare` the type for it. I don't believe we need `declare`, at least where it is currently used. I am also wondering if we can find a better way to export `ISODateFromJson` we export it as a const but it really represents a function.

I agree with the tar.gz changes as well, I am hoping we can hook a GitHub action to deploy the `dist/` from the CI. I also agree with the testing infrastructure for deploys but I am unsure of the approach.

hardy925:
@ChrisDufourMB node deploys are completed, just need the `d.ts` generator to complete that as the types are manual right now

.github/workflows/ts_serialize_release.yml
```diff
@@ -52,13 +52,15 @@ jobs:
...
-          upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps
-          asset_path: ./dist/ts_serialize.tar.gz
-          asset_name: ts_serialize.tar.gz
-          asset_content_type: application/gzip
+          node-version: 12
+          registry-url: https://registry.npmjs.org/
+      - name: deploy project
+        run: |
+          cd dist
+          npm install
```
ChrisDufourMB: Is this `install` call here for generating the package lock file? If it is then we may want to move this into the "build" phase IMO.

ChrisDufourMB (dismissed):
Everything here LGTM, just had a small comment on when we are calling `npm install` when building the npm package.

My only other comment is that renaming the files may make it less clear what they are for if you are importing them from the same directory:

```ts
import { defaultToJson } from "./default.ts";
```
 vs
```ts
import { defaultToJson } from "./default_to_json.ts";
```
On  the other hand the old format has you repeating part of the file name with every inport, and doesn't take advantage of  the directory name :
```ts
import { defaultToJson } from "./to_json/default_to_json.ts";
```


I have no real strong opinion on this either way, so it's fine if you want to use the shorter filenames this way.

PR #57, opened 2020-09-21
Nodejs integration test improvements

ChrisDufourMB:
## Fixes

- Sets up the node example project in a way developers can set up simply by running `npm i && npm test`
- Small improvements to nodeJS integration tests

## Description

Adds some small improvements to the nodeJS integration tests
Improvements to the setup of the nodeJS example project

- Updates `package.json` and adds a `package-lock.json` file to the nodeJS example project
- Small improvements to nodeJS build/test script

## Things to look at

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/examples/node/package.json
+++ b/examples/node/package.json
@@ -7,5 +7,10 @@
     "test": "tsc -p tsconfig.json && node node.js"
   },
   "author": "@gamebridgeai",
-  "license": "ISC"
+  "license": "ISC",
+  "dependencies": {
+    "@gamebridgeai/ts_serialize": "0.2.4",
+    "@types/node": "14.11.1",
+    "typescript": "4.0.3"
+  }
 }
--- a/examples/node/package.json
+++ b/examples/node/package.json
@@ -9,7 +9,7 @@
   "author": "@gamebridgeai",
   "license": "ISC",
   "dependencies": {
-    "@gamebridgeai/ts_serialize": "0.2.4",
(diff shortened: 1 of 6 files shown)


examples/node/package.json
```diff
@@ -7,5 +7,10 @@
     "test": "tsc -p tsconfig.json && node node.js"
   },
   "author": "@gamebridgeai",
-  "license": "ISC"
+  "license": "ISC",
+  "dependencies": {
+    "@gamebridgeai/ts_serialize": "0.2.4",
```
hardy925:
```suggestion
    "@gamebridgeai/ts_serialize": "*",
```

I think this always grabs the latest so we don't need to manually update for releases, maybe we add it for all our packages here.

Thoughts.

hardy925 (commented):
nice!

examples/node/package.json
ChrisDufourMB: Sounds good. We currently don't have a way to automatically update the lockfile, but TBH I don't think it's a big deal b/c we don't need to update the dependency for `ts_serialize`.

The lockfile it's self is ignored unless explicitly added so we shouldn't have any issue with people accidentally committing an updated lockfile. The tests links the build release, so even if the lockfile is pointing to an aincent version of ts_serialize we shouldn't have any problems.

hardy925 (approved):
awesome update

PR #58, opened 2020-09-24
Serialized key function

### LINKED ISSUE #7: [RFC] - Allow property name to take a function in addition to a fixed string
This will allow for programmatic re-naming of property keys such as automatic case conversion.

PizzaCatKing:
## Fixes

fixes #7 

## Description

Adds the ability to use a property key transform function as a serialized key

- Adds ability to pass a function that dynamically transforms the property key into a serialized key

## Things to look at

- [ ] Test coverage
- [ ] Code Style

### DIFF
--- a/serialize_property.ts
+++ b/serialize_property.ts
@@ -12,12 +12,19 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
 export const ERROR_MESSAGE_SYMBOL_PROPERTY_NAME =
   "The key name cannot be inferred from a symbol. A value for serializedName must be provided";
 
+/**
+ * Function to transform a property name into a serialized key programmatically
+ */
+export type ToSerializedKeyStrategy = ((
+  propertyName: string | symbol,
+) => string);
+
 /** string/symbol property name or options for (de)serializing values */
 export type SerializePropertyArgument =
   | string
-  | ((propertyName: string | symbol) => string)
+  | ToSerializedKeyStrategy
   | {
-    serializedKey?: string | ((propertyName: string | symbol) => string);
+    serializedKey?: string | ToSerializedKeyStrategy;
     fromJsonStrategy?:
       | FromJsonStrategy
       | (FromJsonStrategy | FromJsonStrategy[])[];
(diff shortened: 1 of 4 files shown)


serialize_property_test.ts
```diff
@@ -380,3 +380,72 @@ test({
...
   },
 });
+
+test({
+  name: "Serialize key function",
+  fn() {
+    class Test1 extends Serializable {
+      @SerializeProperty(
+        (
+          (propertyName) => "_" + (propertyName as string)
```
PizzaCatKing: This cast is kind of dirty, I think a better way of handling it would be to throw on `Symbol`.

hardy925:
```suggestion
          (propertyName) => `_${String(propertyName)}`
```

I believe this calls `.toString` on Symbol

hardy925:
@PizzaCatKing thoughts on this pr?

hardy925:
hey @PizzaCatKing what is left for this PR? do we need more tests or docs?

serialize_property.ts
```diff
@@ -71,12 +82,19 @@ export function SerializeProperty(
...
+          fromJsonStrategy: decoratorArguments.fromJsonStrategy,
+          toJsonStrategy: decoratorArguments.toJsonStrategy,
+        };
+      } else {
+        decoratorArgumentOptions = {
+          // we can always define serializedKey as decoratorArguments.serializedKey will override this
+          serializedKey: (target as any).tsTransformKey(propertyName),
+          ...decoratorArguments,
+        };
+      }
```
hardy925: what if we abstract setting `decoratorArgumentOptions` to its own function?

ms1111 (approved):
I figure this layers on top of #66, so #66 lets you specify a class-level function for serialized keys and this one lets you specify a property-level function. The order of precedence doesn't change: you can specify a string or function at the property level, and if you don't, the class-level tsTransformKey() will be used.

hardy613 (commented):
I don't think we'll need `ToSerializedKeyStrategy` for node right, 

this is awesome though.

hardy925 (approved):
LGTM! thank you

PR #60, opened 2020-10-31
Fix type exports for deno 1.5.0 where isolatedModules is enabled by default

ms1111:
## Description

deno 1.5.0 enables isolatedModules by default. Compilation was failing with:

`deno test` fails with:
```
error: TS1205 [ERROR]: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.
  FromJsonStrategy,
  ~~~~~~~~~~~~~~~~
    at file:///Users/msmith/git/ts_serialize/mod.ts:11:3

TS1205 [ERROR]: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.
  ToJsonStrategy,
  ~~~~~~~~~~~~~~
    at file:///Users/msmith/git/ts_serialize/mod.ts:12:3

Found 2 errors.
```

FromJsonStrategy and ToJsonStrategy are types. It seems exporting types has to be handled specially when isolatedModules is enabled.

- https://deno.land/manual/getting_started/typescript#code--no-checkcode-option
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-exports

Explicitly export the types with `export type`.

## Testing notes

If you upgrade to 1.5.0, run `deno test` on the develop branch, and hit the error, I find I had to wipe my TypeScript / deno cache (`rm -rf ~/Library/Caches/deno/gen/file/$HOME/git/`) in order for the fix to take effect.

## Things to look at

- [ ] Test coverage
- [ ] Code Style

### DIFF
--- a/mod.ts
+++ b/mod.ts
@@ -5,11 +5,11 @@
 /** property decorator */
 export { SerializeProperty } from "./serialize_property.ts";
 
+export type { FromJsonStrategy, ToJsonStrategy } from "./serializable.ts";
+
 /** abstract class and and compose strategy functions */
 export {
   Serializable,
-  FromJsonStrategy,
-  ToJsonStrategy,
   composeStrategy,
 } from "./serializable.ts";
 


hardy613:
hmmm, I wonder if this is related to the way we make our builds. We keep a manual `ts_serialize.d.ts` file

We should also update the CI scripts for 1.5.0 

- https://github.com/GameBridgeAI/ts_serialize/blob/develop/.github/workflows/ts_serialize_tests.yml#L16
- https://github.com/GameBridgeAI/ts_serialize/blob/develop/.github/workflows/ts_serialize_release.yml#L18
- https://github.com/GameBridgeAI/ts_serialize/blob/develop/.github/workflows/scripts/build.sh#L12

ms1111:
Well, this is odd. Upgrading to v1.5.0 in Github produces a bunch of `deno fmt --check` errors. It wants me to alphabetize a bunch of stuff. But `deno fmt --check` local, also using 1.5.0, says all 23 files are OK.

ms1111:
Oh, I see what's happening. I must have accidentally run `deno fmt` locally and it made changes I didn't commit, so `deno fmt --check` succeeded locally and failed in GitHub.

I'm not super comfortable reformatting the whole project, at least not in this PR; what would you think about having the switch to build with v1.5.0 as a separate PR?

hardy613:
> I'm not super comfortable reformatting the whole project, at least not in this PR; what would you think about having the switch to build with v1.5.0 as a separate PR?

:100: that works for me

ChrisDufourMB:
Personally, I think we should just make one big 1.5.0 upgrade PR, rather than trying to half adapt the current codebase to barely pass the new standards/formatting styles.

For now, we can just work on our "latest supported version" which, (at the time of writing this), is `1.4.0`. You can install this by running `deno upgrade --version=1.4.0` (even if it's "upgrading" to a lower version)

hardy925:
#65 takes care of this be thanks to @ms1111 for pointing it out!

PR #61, opened 2020-10-31
Add JSON types

### LINKED ISSUE #59: [RFC] - JsonNode type
# JsonNode type

add a type for `JsonNode` that consists of the correct types supported by the [JSON Notation standard](https://tools.ietf.org/html/rfc8259) 

[text (see "3. Values")](https://tools.ietf.org/rfc/rfc8259.txt)

Note: if updated please follow the `Obsoleted by:` tag in the header: 
Screen Shot 2020-10-16 at 5 42 04 PM

## Motivation
To properly type our strategic directions.

`ToJsonStrategy`  and `FromJsonStrategy` make use of `any`, by providing this type we can set up directional strategies.

## Design Detail

The supported RFC types should be `|` (Unioned) into a `JsonNode` type 

```ts
/** Functions used when hydrating data */
export type FromJsonStrategy = (value: any) => any;

/** Functions used when dehydrating data */
export type ToJsonStrategy = (value: any) => any;
```

Becomes

```ts
/** Functions used when hydrating data */
export type FromJsonStrategy<T> = (value: JsonNode) => T;

/** Functions used when dehydrating data */
export type ToJsonStrategy <T>= (value: T) => JsonNode;
```

ms1111:
## Fixes

#59 - JsonNode type

## Description

Define explicit types for legal plain JSON values to avoid double-deserialization shenanigans.

Fixes #59.

Adds JsonValue and JsonObject types to make ToJsonStrategy and FromJsonStrategy more specific.

## Things to look at

- [ ] Test coverage
- [ ] Code Style

### DIFF
--- a/from_json/as.ts
+++ b/from_json/as.ts
@@ -6,7 +6,7 @@ import { FromJsonStrategy, Serializable } from "../serializable.ts";
 export function fromJsonAs<T>(
   type: T & { new (): Serializable },
 ): FromJsonStrategy {
-  return value => {
+  return (value) => {
     if (Array.isArray(value)) {
       return value.map((item) => new type().fromJson(item));
     }
--- a/from_json/as.ts
+++ b/from_json/as.ts
@@ -11,10 +11,10 @@ import {
 export function fromJsonAs<T>(
   type: T & { new (): Serializable },
 ): FromJsonStrategy {
-  return (value: T) => {
+  return (value: JsonValue) => {
     if (Array.isArray(value)) {
-      return value.map((item) => new type().fromJson(item));
+      return value.map((item) => new type().fromJson(item as JsonObject));
     }
-    return new type().fromJson(value);
+    return new type().fromJson(value as JsonObject);
   };
 }
--- a/from_json/as.ts
+++ b/from_json/as.ts
@@ -2,8 +2,6 @@
 
 import {
   FromJsonStrategy,
-  JsonObject,
-  JsonValue,
   Serializable,
 } from "../serializable.ts";
 
@@ -11,10 +9,10 @@ import {
 export function fromJsonAs<T>(
   type: T & { new (): Serializable },
 ): FromJsonStrategy {
-  return (value: JsonValue) => {
+  return value => {
     if (Array.isArray(value)) {
-      return value.map((item) => new type().fromJson(item as JsonObject));
+      return value.map((item) => new type().fromJson(item));
     }
-    return new type().fromJson(value as JsonObject);
+    return new type().fromJson(value);
   };
 }
--- a/from_json/as.ts
+++ b/from_json/as.ts
@@ -1,9 +1,6 @@
 // Copyright 2018-2020 Gamebridge.ai authors. All rights reserved. MIT license.
 
-import {
-  FromJsonStrategy,
-  Serializable,
-} from "../serializable.ts";
+import { FromJsonStrategy, Serializable } from "../serializable.ts";
 
 /** revive data using `fromJson` on a subclass type */
 export function fromJsonAs<T>(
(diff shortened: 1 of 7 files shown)


from_json/as.ts
```diff
@@ -1,10 +1,10 @@
...
 
-import { FromJsonStrategy, Serializable } from "../serializable.ts";
+import { FromJsonStrategy, JsonObject, JsonValue, Serializable } from "../serializable.ts";
 
 /** revive data using `fromJson` on a subclass type */
 export function fromJsonAs<T>(
   type: T & { new (): Serializable },
 ): FromJsonStrategy {
-  return (value: T) => new type().fromJson(value);
+  return (value: JsonValue) => new type().fromJson(value as JsonObject);
```
ms1111: Having to cast `value as JsonObject` here makes me think maybe I'm doing something wrong, or having the separate distinction between JsonValue and JsonObject isn't as helpful as I'd hoped.

`fromJsonAs()` wants to convert a JsonObject to a user-defined type, but FromJsonStrategy can't be specific about JsonObject since it's also used for field-level deserialization. So the arrow function has to take a JsonValue, and yet we want to pass a nested JsonObject to fromJson().

serialize_property_test.ts
```diff
@@ -216,7 +216,7 @@ test({
     }
     class Test extends Serializable {
       @SerializeProperty({
-        fromJsonStrategy: (v: OtherClass) => new OtherClass().fromJson(v),
+        fromJsonStrategy: (v: JsonValue) => new OtherClass().fromJson(v as JsonObject),
```
ms1111: ... and here, too

serializable.ts
```diff
@@ -4,24 +4,36 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
...
 export abstract class Serializable {
   public toJson(): string {
     return toJson(this);
   }
   public fromJson(): this;
   public fromJson(json: string): this;
-  public fromJson(json: Record<string, any>): this;
-  public fromJson(json: string | Record<string, any> = {}): this {
+  public fromJson(json: JsonObject): this;
+  public fromJson(json: string | JsonObject = {}): this {
```
hardy613: I really wonder if the previous `Record<>`'s were a bug in disguise

hardy613 (commented):
I will probably pull this down to look into some things, but this is a great start! thanks @ms1111

serialize_property_test.ts
```diff
@@ -348,7 +349,8 @@ test({
     class Test2 extends Serializable {
       @SerializeProperty({
         serializedKey: "serialize_me_2",
-        fromJsonStrategy: (json) => new Test1().fromJson(json),
+        fromJsonStrategy: (json: JsonValue) =>
```
ChrisDufourMB: Typing here is implied by the `SerializePropertyArgument` -> `FromJsonStrategy` types, no need to manually specify it
```suggestion
        fromJsonStrategy: (json) =>
```

Screen Shot 2020-11-03 at 20 00 13

serializable.ts
ChrisDufourMB: Unrelated note:

I think we should remove the default `{}` value and the `fromJson()` (no args) interface. IMO it doesn't really make sense to be able to serialize nothing into something

ChrisDufourMB: Thoughts on using `JsonValue` here instead? We currently don't have the ability to set the JSON root property of a class key, but being able to deserialize a pure value, or a JSON array may be useful in some contexts.

Implementing this may be a bunch of work on the de-serializer side of things though

ChrisDufourMB (commented):
Looks great! Some comments around JsonValue vs JsonObject

hardy925 (dismissed):
Thanks @ms1111 LGTM

hardy925 (approved):

PR #66, opened 2020-11-04
tsTransformKey functionality

### LINKED ISSUE #29: [RFC] - snake_case and camelCase conversions
# Auto snake_case and camelCase conversions

Have a setting or API for setting properties to be auto snake_cased or camelCased.

## Motivation

Time saver without the risk of overriding `content-type`. relates to #7 

## Design Detail

I was thinking of a setting that could be defined in the classes that extend `Serializable`, then if a base class extends and uses the settings to turn on auto-casing then classes that extend the base class would inherit that setting. These classes could optionally turn them off per class or per property with a new `SerializeProperty` option.

hardy925:
## Fixes

closes #29 

## Description

Adds optional `tsTransformKey` functionality

- tsTransformKey to ... transform keys
- is inherited by children
- children can provide their own to override
- serialize property options override the transforms
- tests
- docs

ps. do we need a website now?

## Things to look at

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
+++ b/.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
@@ -1,6 +1,16 @@
 declare module "@gamebridgeai/ts_serialize" {
+  /** to be implemented by external authors on their models  */
+  export interface TransformKey {
+    /** a function that will be called against
+   * every property key transforming the key
+   * with the provided function
+   */
+    tsTransformKey(key: string): string;
+  }
+
   /** Adds methods for serialization */
   export abstract class Serializable {
+    public tsTransformKey?(key: string): string;
     public toJson(): string;
     public fromJson(): this;
     public fromJson(json: string): this;
--- a/serializable.ts
+++ b/serializable.ts
@@ -4,8 +4,15 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
 import { defaultToJson } from "./to_json/default.ts";
 import { recursiveToJson } from "./to_json/recursive.ts";
 
+export declare interface TransformKey {
+  tsTransformKey(key: string): string;
+}
+
 /** Adds methods for serialization */
 export abstract class Serializable {
+  public tsTransformKey?(key: string): string {
+    return key;
+  }
   public toJson(): string {
     return toJson(this);
   }
@@ -80,8 +87,8 @@ const ERROR_MESSAGE_MISSING_PROPERTIES_MAP =
   "Unable to load serializer properties for the given context";
 
 /** Converts to object using mapped keys */
-export function toPojo<T>(
-  context: Record<keyof T, unknown>,
+export function toPojo(
+  context: any,
 ): Record<string, unknown> {
   const serializablePropertyMap = SERIALIZABLE_CLASS_MAP.get(
     context?.constructor?.prototype,
@@ -102,7 +109,7 @@ export function toPojo<T>(
     } of serializablePropertyMap.propertyOptions()
   ) {
     // Assume that key is always a string, a check is done earlier in SerializeProperty
-    const value = context[propertyKey as keyof T];
+    const value = context[String(propertyKey)];
 
     // If the value is serializable then use the recursive replacer
     if (
--- a/serializable.ts
+++ b/serializable.ts
@@ -4,12 +4,18 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
 import { defaultToJson } from "./to_json/default.ts";
 import { recursiveToJson } from "./to_json/recursive.ts";
 
+/** to be implemented by external authors on their models  */
 export declare interface TransformKey {
+  /** a function that will be called against
+   * every property key transforming the key
+   * with the provided function
+   */
   tsTransformKey(key: string): string;
 }
 
 /** Adds methods for serialization */
 export abstract class Serializable {
+  /** the default transform functionality */
   public tsTransformKey?(key: string): string {
     return key;
   }
--- a/serializable.ts
+++ b/serializable.ts
@@ -109,7 +109,7 @@ export function toPojo(
     } of serializablePropertyMap.propertyOptions()
   ) {
     // Assume that key is always a string, a check is done earlier in SerializeProperty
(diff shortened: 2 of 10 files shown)


serialize_property.ts
```diff
@@ -72,8 +72,13 @@ export function SerializeProperty(
         throw new Error(ERROR_MESSAGE_SYMBOL_PROPERTY_NAME);
       }
 
+      const transformKeyFunction =
+        Object.prototype.hasOwnProperty.call(target, "tsTransformKey")
+          ? target.tsTransformKey
+          : Object.getPrototypeOf(target).tsTransformKey;
+
```
hardy925: so this will happen for every property, thoughts?

serialize_property.ts
```diff
@@ -72,8 +72,14 @@ export function SerializeProperty(
         throw new Error(ERROR_MESSAGE_SYMBOL_PROPERTY_NAME);
       }
 
+      // get current class or parent class function
+      const transformKeyFunction =
+        Object.prototype.hasOwnProperty.call(target, "tsTransformKey")
+          ? (target as any).tsTransformKey
+          : Object.getPrototypeOf(target).tsTransformKey;
```
ms1111: Would just `const transformKeyFunction = target.tsTransformKey` work?

.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
```diff
@@ -1,6 +1,16 @@
...
+    /** a function that will be called against
+   * every property key transforming the key
+   * with the provided function
+   */
+    tsTransformKey(key: string): string;
+  }
+
   /** Adds methods for serialization */
   export abstract class Serializable {
+    public tsTransformKey?(key: string): string;
```
ms1111, line 13: What would you think about having a decorator on the class instead of a method?

```
@SerializeDefaults({ transformKey: TransformKey })
class Foo extends Serializable { ... }
```

... I looked into decorator inheritance a bit and it might be more trouble than it's worth, but just putting it out there.

https://medium.com/@ttemplier/angular2-decorators-and-class-inheritance-905921dbd1b7

serialize_property.ts
hardy925: yes, thank youu!

.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
hardy925: oh boy, JS decorators are so wonky, class decorators run after the property decorators, so this would require a big refactor. I believe class decorators are called when the class is initialized where property decorators are called during the construction of the class. 

With how we make our map we would need to update the map on the fly when the class is used and I am not sold on doing that, we did try a class decorator in an earlier iteration of the project and quickly ran into issues. 

I also prefer the approach in the PR using a class function and interface. We add the interface (and declare it) too. get warnings on. implementation. As the function is optional we need the interface for these warnings. That said authors can incorrectly have a tsTransformKey function signature without errors. if it is not implemented in the class, but this is okay IMHO because we provide the interface with instructions on how to implement it. `(s:  string): string`

serializable.ts
```diff
@@ -4,8 +4,21 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
 import { defaultToJson } from "./to_json/default.ts";
 import { recursiveToJson } from "./to_json/recursive.ts";
 
+/** to be implemented by external authors on their models  */
+export declare interface TransformKey {
```
ms1111, line 8: Is the “declare” keyword needed here?

.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
```diff
@@ -1,6 +1,16 @@
 declare module "@gamebridgeai/ts_serialize" {
+  /** to be implemented by external authors on their models  */
+  export interface TransformKey {
+    /** a function that will be called against
+   * every property key transforming the key
+   * with the provided function
+   */
+    tsTransformKey(key: string): string;
```
ms1111, line 8: Maybe we need two - one for ser and one to deser. (Camel to snake vs snake to camel)

hardy613: We always have the original property key so the transform only needs to be one way.

serializable.ts
hardy613: yea, without it we do not get the errors to populate if implemented wrong when declared with `implements TransformKey`

Although some more testing around this wouldn't hurt

ms1111 (approved):

serializable.ts
```diff
@@ -4,8 +4,21 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
...
+  /** a function that will be called against
+   * every property key transforming the key
+   * with the provided function
+   */
+  tsTransformKey(key: string): string;
```
ChrisDufourMB, line 13: It's possible to use a symbol here if we want maximum collision avoidance:
```ts
export const tsTransformKey: unique symbol = Symbol("tsTransformKey");

export declare interface TransformKey {
  [tsTransformKey](key: string): string;
}

export abstract class Serializable {
  public [tsTransformKey]?(key: string): string {
    return key;
  }
  ...
}

// Implementation
class TestTransformKey extends Serializable implements TransformKey {
  public [tsTransformKey](key: string): string {
    return `__${key}__`;
  }
}
```

This is probably way overkill though and would be annoying to use in practice

serializable.ts
```diff
@@ -4,8 +4,21 @@ import { SerializePropertyOptionsMap } from "./serialize_property_options_map.ts
...
+}
+
 /** Adds methods for serialization */
 export abstract class Serializable {
+  /** the default transform functionality */
+  public tsTransformKey?(key: string): string {
```
ChrisDufourMB, line 19: I am a little sad we can't make this `static` due to typescript limitations, at least for now https://github.com/microsoft/TypeScript/issues/33892

ChrisDufourMB (approved):
💯

serializable.ts
hardy613: i think we'd hit an issue with this when users go to implement their own they would also need to export the symbol and use it accordingly, I think its a bit much to ask of external authors.

PR #79, opened 2020-11-18
date updates

hardy925:
## Fixes

_Please link to the issue this PR addresses_

## Description

better ISO 8106 support

## Things to look at

- [x] Test coverage
- [x] Code Style


from_json/date.ts
```diff
@@ -11,13 +11,78 @@ export function createDateStrategy(regex: RegExp): FromJsonStrategy {
...
+}
+
+/** Test a string for a full ISO 8601 Date */
+export function iso8601Date(input: JsonValue): any {
+  var iso =
+    /^(\d{4})(?:-?W(\d+)(?:-?(\d+)D?)?|(?:-(\d+))?-(\d+))(?:[T ](\d+):(\d+)(?::(\d+)(?:\.(\d+))?)?)?(?:Z(-?\d*))?$/;
+  // throw error on bad data?
+  if (typeof input !== "string" || !iso.exec(input)) {
+    return input;
+  }
```
ChrisDufourMB: I think we should throw an error here instead of returning the input, or at least return `undefined`.

If we eventually add pre-parse validation we should have a proper way of validating this and handling errors, but for now, I think we should just try to avoid returning different types when errors occur.

Another option would be to simply return an `Invalid Date` type date, not sure how we would initialize that though.

from_json/date.ts
```diff
@@ -11,13 +11,78 @@ export function createDateStrategy(regex: RegExp): FromJsonStrategy {
...
+  year: number;
+  week: number;
+  weekday: number;
+  timezone: number;
+}
+
+/** Test a string for a full ISO 8601 Date */
+export function iso8601Date(input: JsonValue): any {
+  var iso =
+    /^(\d{4})(?:-?W(\d+)(?:-?(\d+)D?)?|(?:-(\d+))?-(\d+))(?:[T ](\d+):(\d+)(?::(\d+)(?:\.(\d+))?)?)?(?:Z(-?\d*))?$/;
```
ChrisDufourMB: A comment explaining what this regex does would be appreciated

from_json/date.ts
```diff
@@ -11,13 +11,78 @@ export function createDateStrategy(regex: RegExp): FromJsonStrategy {
...
+  // throw error on bad data?
+  if (typeof input !== "string" || !iso.exec(input)) {
+    return input;
+  }
+  // first match is the whole pattern which we do not need `,` skips it
+  // we're also skipping week and weekday
+  let [
```
ChrisDufourMB: Could you add a quick comment with an example array? There are a lot of unused variables, so it's a bit difficult to understand what the format coming out of `input.match(iso)` would be like

from_json/date.ts
```diff
@@ -11,13 +11,78 @@ export function createDateStrategy(regex: RegExp): FromJsonStrategy {
...
+      year,
+      ,
+      ,
+      month,
+      day,
+      hour,
+      minute,
+      second,
+      millisecond,
+      zone,
```
ChrisDufourMB: How does the zone support work? IIRC zones can be something like this:`+02:30`, which can't be cleanly converted to a number

Also note that a zone like this: `Z+0930` is not + 930 hours, but +9 hours, 30 minutes (basically it's just 09:30 without the `:`)

from_json/date.ts
```diff
@@ -11,13 +11,78 @@ export function createDateStrategy(regex: RegExp): FromJsonStrategy {
...
+    ] = input.match(iso)?.map(Number) || [],
+    timezone = parseTimezone(0);
+
+  if (Number.isNaN(year)) {
+    return input;
```
ChrisDufourMB: See other comment on error handling, same applies here

from_json/date.ts
```diff
@@ -11,13 +11,78 @@ export function createDateStrategy(regex: RegExp): FromJsonStrategy {
...
- *  YYYY-MM-DDThh:mm:ss.sTZD
- * @see: https://www.w3.org/TR/NOTE-datetime
- * @type {RegExp}
- */
-export const ISODateFromJson = createDateStrategy(
-  /^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d(\.\d+)?(([+-]\d\d:\d\d)|Z)?$/i,
-);
+/** converts timezone to minutes */
+function parseTimezone(zone: number): number {
+  return zone * 60 - Number.parseInt(`-${new Date().getTimezoneOffset()}`);
```
ChrisDufourMB: `new Date().getTimezoneOffset()` returns a number, it seems a bit strange to me that we would be converting it to a string, then parsing it back to a number like this. Parsing it as a number, then subtracting it seems a bit redundant, would it make more sense to simply add it instead?

Does this work in negative time zones?

from_json/date_test.ts
```diff
@@ -1,72 +1,81 @@
...
-    assertEquals(testObj.date.getFullYear(), 2020);
-    assert(!(testObj.not_a_date instanceof Date));
-    assertEquals(testObj.not_a_date, "Hello world");
+    assertEquals(testObj.date.toISOString(), "2020-12-31T12:00:00.000Z");
   },
 });
 
 test({
-  name: "ISODateFromJson parses ISO dates - 2020-12-31T23:00:00",
+  name: "iso8601Date parses with added timezone - 2020-12-31T12:00:00Z-6",
```
ChrisDufourMB: I don't think single-digit offsets are technically part of the ISO standard, I believe that a proper year offset needs to be left-padded with 0s

ChrisDufourMB (commented):
This doesn't appear to support all the different kinds of timezones that could be part of an ISO date, what degree of support are you looking to implement? I feel like we will need more tests cases for all the different variants and edge cases that can happen

Here is the test I tried, for reference
```ts

test({
  name: "iso8601Date parses with added timezone - 2020-12-31T12:00:00Z-06:00",
  fn() {
    const testJson = `{"date":"2020-12-31T06:00:00Z-06:00"}`;
    const testObj = JSON.parse(testJson, (_, v) => iso8601Date(v));
    assert(testObj.date instanceof Date);
    assertEquals(testObj.date.toISOString(), "2020-12-31T12:00:00.000Z-06:00");
  },
});
```

from_json/date.ts
ms1111: Throwing an Error would be kind of how Jackson handles invalid input in Java.

from_json/date_test.ts
```diff
@@ -1,72 +1,81 @@
...
+});
+
+test({
+  name: "iso8601Date parses `2020-366` - leap year",
+  fn() {
+    const testJson = `{"date":"2020-366"}`;
+    const testObj = JSON.parse(testJson, (_, v) => iso8601Date(v));
+    assert(testObj.date instanceof Date);
+    assertEquals(testObj.date.toISOString(), "2020-12-31T00:00:00.000Z");
   },
```
ms1111, line 84: Can I suggest a test to check that YYYY-MM-DD is deserialized as midnight local time? (I think this test will pass, looking at the code)

This is how we'll want it for our date types. (Code that uses isoDateStringToLocalDate() now)

Here is how we tested it in date-utils.spec.ts:
```
    // 1990-12-11
    const localDateMidnight = new Date(1990, 11, 11);
...
expect(isoDateStringToLocalDate("1990-12-11").getTime()).toEqual(localDateMidnight.getTime());
```

ms1111 (commented):
Would you also want to have a serializer in the library to render a local date as an ISO date string (YYYY-MM-DD) in toJson()? Or would you want apps to provide that themselves (like our localDateToIsoDateString() function)

Here's the test case for localDateToIsoDateString() btw:
```
    // 1990-12-11
    const localDateMidnight = new Date(1990, 11, 11);

expect(localDateToIsoDateString(localDateMidnight)).toEqual("1990-12-11");
```

code:
```
export function localDateToIsoDateString(date: Date): string {
    const year = date.getFullYear();
    const month = date.getMonth() + 1;
    const day = date.getDate();
    return `${formatISOYear(year)}-${padLeft(month, 2)}-${padLeft(day, 2)}`;
}
```

from_json/date.ts
hardy925: Removing this in place for `Date.UTC`

from_json/date.ts
hardy925: done

from_json/date.ts
hardy925: remove most of this, and I used `match` ... exec would be better for this

from_json/date.ts
hardy925: done

from_json/date_test.ts
hardy925: This depends on the string, "1990-12-11" will have zones applied to it,  where "1990/12/11" will not. I am going to divert this so it is up to authors to make the strategies they want now.

hardy925:
@ms1111 I think we'll go the route of having authors write their strategies, the two we provide for dates is more than enough IMHO.

PR #84, opened 2020-11-26
added tsSerialize()  exposing toPojo, clean up types, rename Json -> JSON program wide

### LINKED ISSUE #63: [RFC] - Easier Angular HttpClient serialization
# Easier Angular HttpClient serialization

Expose toPojo() in Serializable or make toJson() return a JsonObject (related to #59).

## Motivation

In Angular, if you do something like:
```ts
return this.httpClient.post(this.url, mySerializableObject.toJson());
```
... it tries to send the JSON string as `Content-type: text/plain` instead of `application/json`.

## Design Detail
Add a method to the Serializable class expose the internal serialized representation (the POJO before `JSON.stringify()`). In Angular you could then do:
```ts
return this.httpClient.post(this.url, mySerializableObject.toPojo());
```

## Drawbacks
- more noise/confusion in the API

## Rationale and Alternatives
Alternatives:
1. Some kind of HttpClient intereptor to detect a Serializable object and set the character set appropriately. The caller would then do:
```ts
return this.httpClient.post(this.url, mySerializableObject); // some magic will call toJson() on the object
```

2. Or, callers could set HttpClient options to force the right content type on each post/put call.
```ts
const httpOptions = {
    headers: new HttpHeaders({
        'Content-Type': 'application/json',
    })
};
return this.httpClient.post(this.url, mySerializableObject.toJson(), httpOptions);
```

### LINKED ISSUE #81: [RFC] - type clean up
deno doc. types are a mess

hardy925:
## Fixes

fixes #81 
fixes #63 

## Description
better deno docs, and fixing some RFCs

- new function tsSeriailize exposing toPojo
- clean up types for deno docs
- using JS naming convention for JSON

## Things to look at

- [x] Test coverage
- [x] Code Style


.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
```diff
@@ -1,97 +1,89 @@
...
-      | ToJsonStrategyArgument
-  ): FromJsonStrategy | ToJsonStrategy;
+      | FromJSONStrategyArgument
+      | ToJSONStrategyArgument
+  ): FromJSONStrategy | ToJSONStrategy;
 
-  /** revive data using `fromJson` on a subclass type */
-  export function fromJsonAs<T>(
+  /** revive data using `fromJSON` on a subclass type */
+  export function as<T>(
```
ms1111: I worry a bit about polluting the caller's namespace with a generic name like `as()`, it seems kind of like rxjs's operator names.

ms1111 (dismissed):
Looks good. I have a comment about `as()`.

.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
hardy925: > it seems kind of like rxjs's

ya it is silly because I was going to comment with "oh they can just change via the `as` keyword"  so now I am thinking

error_messages.ts
```diff
@@ -0,0 +1,13 @@
...
+  "Unable to load serializer properties for the given context";
+
+export const ERROR_MESSAGE_DUPLICATE_PROPERTY_KEY =
+  "This key name is already in use by another property, please use a different name";
+
+export const ERROR_MESSAGE_DUPLICATE_SERIALIZE_KEY =
+  "This serialize key is already in use by another property, please use a different name";
+
+export const ERROR_MESSAGE_SYMBOL_PROPERTY_NAME =
+  "The key name cannot be inferred from a symbol. A value for serializedName must be provided";
```
ChrisDufourMB, line 13: Should we export these are part of `mod.ts`? That way implementors can catch the specific errors if they occur by comparing these against the error's message

strategy/compose_strategy.ts
```diff
@@ -0,0 +1,26 @@
...
+/** Functions used when dehydrating data */
+export type ToJSONStrategy = (value: any) => JSONValue;
+export type ToJSONStrategyArgument = (ToJSONStrategy | ToJSONStrategy[])[];
+/** Function to build a `fromJSONStrategy` or `toJSONStrategy`.
+ * Converts value from functions provided as parameters
+ */
+export function composeStrategy(
+  ...fns:
+    | FromJSONStrategyArgument
+    | ToJSONStrategyArgument
```
ChrisDufourMB, line 18: The way this is written you could technically chain both `from` and `to` strategies together
```ts
composeStrategy(toJSONDefault, fromJSONDefault);
```

Not sure if this would be an issue as I imagine this would mostly be used for values that are subclasses of both anyways (like `string` -> `string`)

ChrisDufourMB (dismissed):
Awesome work! 💯

hardy925:
@ChrisDufourMB nodejs tests are failing on the polymorphism changes

Not sure whats different thought, deno is able to understand the context switch

Screen Shot 2020-11-27 at 3 11 05 AM

Screen Shot 2020-11-27 at 3 11 12 AM

error_messages.ts
hardy925: hm, idk do you think it would get used somehow?

strategy/compose_strategy.ts
hardy925: ya, It's okay, the way it was typed before was basically the same thing

PR #85, opened 2020-11-26
Fix empty class serialization

### LINKED ISSUE #83: [BUG] - Unable to serialize/de-serialize classes without any serializable properties
**Describe the bug**

**To Reproduce**
- Create the following class:
```ts
class TestSerializable extends Serializable {}
```
- Try to serialize it/deserialize an json object into it
- Serialization will fail with the following error:
`Unable to load serializer properties for the given context: [object Object]`

**Expected behaviour**
The class should serialize to `{}` and inherit values from it's parents

- OS: macOS Catalina
- Browser: `deno`
- Version: 1.5.4

**Additional context**

ChrisDufourMB:
## Fixes

fixes #83 

## Description

Fixes issue where classes without any `@SerializeProperty()` decorator fails to be serialized

- Add logic to create serialization definitions on class construction if they weren't created when a `SerializeProperty` decorator was processed. This is recursive as parent constructors aren't called unless explicitly called via `super()`

## Things to look at

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/serializable.ts
+++ b/serializable.ts
@@ -45,6 +45,52 @@ export abstract class Serializable {
   public fromJson(json: JsonValue | Object): this {
     return fromJson(this, json);
   }
+
+  constructor() {
+    this.getOrInitializeDefaultSerializerLogicForParents();
+  }
+
+  // Recursively set default serializer logic for own class definition and parent definitions if none exists
+  private getOrInitializeDefaultSerializerLogicForParents(
+    targetPrototype = this.constructor.prototype, // This will the the class instance, regardless of how low level it is
+  ): SerializePropertyOptionsMap | undefined {
+    // Don't create serialization logic for Serializable
+    if (targetPrototype === Serializable.prototype) {
+      return undefined;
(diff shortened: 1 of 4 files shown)


ms1111 (approved):
Looks sane

serializable.ts
```diff
@@ -45,6 +45,52 @@ export abstract class Serializable {
...
+      SERIALIZABLE_CLASS_MAP.set(
+        targetPrototype,
+        new SerializePropertyOptionsMap(parentMap),
+      );
+      const newSerializableOptions = SERIALIZABLE_CLASS_MAP.get(
+        targetPrototype,
+      );
+
+      if (!newSerializableOptions) {
+        // This shouldn't happen considering we just added that value
```
hardy925, line 85: I am honoured we found a way to have a legit `This shouldn't happen` comment!!!!

:))))))))

hardy925 (approved):
🔥 🧯

PR #86, opened 2020-11-27
Turn on bash strict mode for release scripts

ms1111:
## Description
Exit early if there's an error to make it easier to spot.

When I forgot to remove `dist` before running `.github/workflows/scripts/build.sh v0.5.0` an error was reported early but didn't abort the script, so it hit other errors later.

Now it'll stop when it hits an error.

## Things to look at

- [ ] Test coverage
- [ ] Code Style


hardy925 (approved):
nice

PR #90, opened 2020-12-08
fixes #89 added `toObjectContaining`

### LINKED ISSUE #89: [RFC] - Built-in fromJSONStrategy for objects
Include a function in the library to better support fields like `someField: { [key: string]: SomeClass }`

**Motivation**

I see something like this fairly often:
```ts
    @SerializeProperty({
        fromJSONStrategy: v => {
            const record = {};
            for (const key in v as {}) {
                if (v[key]) {
                    record[key] = new SomeClass().fromJSON(v[key]);
                }
            }
            return record;
        },
    })
    public someField: { [key in SomeEnum]?: SomeClass };
```

For arrays, we handle this out of the box with `fromJsonStrategy: toSerialize(constructor)`. For example, this works:

```ts
    @SerializeProperty({ fromJSONStrategy: toSerialize(SomeClass) })
    public someField: SomeClass[];
```

It would be nice to get to something more concise, like:

```ts
    @SerializeProperty({ fromJsonStrategy: deserializeMapValues(SomeClass) })
    public someMap: { [k: string]: SomeClass };
```

Nicer still if it supports maps of arrays:
```ts
    @SerializeProperty({ fromJsonStrategy: deserializeMapValues(SomeClass) })
    public someMap: { [k: string]: SomeClass[] };
```

**Design Detail**

I believe we could export a generic `deserializeMapValues<T>(new () => T)` that works similarly to the specific code at the top of the description.

It would work similarly to `toSerialize()`.

**Drawbacks**

It would be too good

**Unresolved questions**

hardy925:
**Description**

fixes #89 adding `toObjectContaining` to revive to

```ts
{ [k: string]: SomeClass }
// or
{ [k: string]: SomeClass[] }
```

- add `toObjectContaining`
- added tests for `toObjectContaining`
- add docs for `toObjectContaining`

**Things to look at**

- [x] Test coverage
- [x] Code Style


strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,30 @@
...
+            record[prop] = (value[prop] as JSONArray).map((v: JSONValue) =>
+              new type().fromJSON(v)
+            );
+          } else {
+            record[prop] = new type().fromJSON(value[prop]);
+          }
+        }
+      }
+    }
+    return record;
```
ms1111, line 56: What should we do if it's not null/undefined, but is not an object here? (A string or a number.) Should we throw an error or let it through?

ms1111 (dismissed):
Nice!

strategy/from_json/to_object_containing.ts
hardy925: added errors!, thanks!

strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,62 @@
...
+import { FromJSONStrategy } from "../compose_strategy.ts";
+import {
+  ERROR_TO_OBJECT_CONTAINING_INVALID_SUB_VALUE,
+  ERROR_TO_OBJECT_CONTAINING_INVALID_VALUE,
+  ERROR_TO_OBJECT_CONTAINING_USE_TO_SERIALIZABLE,
+} from "../../error_messages.ts";
+
+/** revive data from `{k: v}` using `fromJSON` on a subclass type `v` */
+export function toObjectContaining<T>(
+  type: T & { new (): Serializable },
```
ChrisDufourMB: I am a bit concerned with the `{ new (): Serializable }` type here as it forces the Object class to not implement a constructor with arguments, which may not be the case for some classes.

image
^ `toObjectContaining` won't accept `SomeClass` as an argument becase it doesn't have a no args constructor

I think an initializer function could work around this:
```ts
toObjectContaining(() => new SomeClass("defaultValue"))
```

strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,62 @@
...
+  ) {
+    if (value == null) {
+      return null;
+    }
+
+    if (Array.isArray(value)) {
+      throw Error(ERROR_TO_OBJECT_CONTAINING_USE_TO_SERIALIZABLE);
+    }
+
+    if (Object.prototype.toString.call(value) !== "[object Object]") {
```
ChrisDufourMB: Is this the best way to check is a given value is a POJO object? Maybe a comment here could help.

strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw Error(ERROR_TO_OBJECT_CONTAINING_INVALID_VALUE);
+    }
+
+    const record: Record<string, Serializable | Serializable[] | null> = {};
+    // check that JSONValue is something we can deal with
+    // but mostly to make the type checker happy
+    if (typeof value === "object" && !Array.isArray(value)) {
```
ChrisDufourMB, line 29:
```suggestion
    if (typeof value === "object") {
```
The earlier check should have caught any arrays

strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,62 @@
...
+          record[prop] = null;
+          continue;
+        }
+
+        if (Array.isArray(value[prop])) {
+          record[prop] = (value[prop] as JSONArray).map((v: JSONValue) => {
+            if (
+              Object.prototype.toString.call(v) !== "[object Object]"
+            ) {
+              throw Error(ERROR_TO_OBJECT_CONTAINING_INVALID_SUB_VALUE);
```
ChrisDufourMB:
```suggestion
              throw new Error(ERROR_TO_OBJECT_CONTAINING_INVALID_SUB_VALUE);
```

Looks like there is a missing `new` keyword?

strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,62 @@
...
+            }
+            return new type().fromJSON(v);
+          });
+          continue;
+        }
+
+        if (Object.prototype.toString.call(value[prop]) !== "[object Object]") {
```
ChrisDufourMB: This pattern is used in a few places, maybe a util function would be useful?

ChrisDufourMB (dismissed):
Blocking because `serializable.ts` changes break handling empty arrays

The `toObjectContaining` work looks good but enforces a specific constructor pattern which may make implementing it difficult in some circumstances

hardy925:
> Blocking because serializable.ts changes break handling empty arrays

Am I missing a test?

serializable.ts
```diff
@@ -193,7 +193,7 @@ function fromJSON<T>(
           fromJSONStrategy = fromJSONDefault,
         } = serializablePropertyMap.getBySerializedKey(key) || {};
 
-        const processedValue: unknown = Array.isArray(value)
+        const processedValue: unknown = Array.isArray(value) && value.length
```
ChrisDufourMB: @shardyMBAI Sorry, seems that this review comment didn't actually get added somehow

Adding in this constraint causes empty arrays to not be serialized into arrays:
```
test({
  name: "Revives an empty array of `type`",
  fn() {
    class OtherClass extends Serializable {
      @SerializeProperty()
      id!: number;
    }
      @SerializeProperty({
        fromJSONStrategy: (v) => new OtherClass().fromJSON(v),
      })
      array!: OtherClass[];
    }
    const testObj = new Test().fromJSON(
      `{"array":[]}`,
    );

    assert(Array.isArray(testObj.array)); // Test fails here - Test { array: OtherClass {} }
    assertEquals(testObj.array.length, 0);
  },
});
```

hardy925: damnit, thanks!

strategy/from_json/date.ts
```diff
@@ -1,12 +1,13 @@
...
 import { FromJSONStrategy } from "../compose_strategy.ts";
 import { JSONValue } from "../../serializable.ts";
+import { ERROR_INVALID_DATE } from "../../error_messages.ts";
 
 /** allows authors to pass a regex to parse as a date */
 export function createDateStrategy(regex: RegExp): FromJSONStrategy {
   return function _createDateStrategy(input: JSONValue): Date {
     if (typeof input !== "string" || !regex.exec(input)) {
-      throw new Error("invalid date");
+      throw new Error(ERROR_INVALID_DATE);
```
WillWang-MindBridge, line 10: I see we are only using the passed in regex for validation. Should you extract the 3 numbers and use that to build the date instead?

hardy925: Hey, 
There are probably some improvements that could be made here, but the goal is to parse a string and convert that to a Date if it passes. We're not sure what string or regex we'll get as well. Is there a bug it would be fixing?

WillWang-MindBridge: regex: `^Im going to shoot my foot$`
input: `Im going to shoot my foot`
it's gonna fail at `new Date(input)`
I agree it's not a main concern

strategy/from_json/to_object_containing.ts
hardy925: oh ya, added the above check pretty late, thanks

strategy/from_json/to_object_containing.ts
hardy925: Yea it is, I know it's expensive

strategy/from_json/to_object_containing.ts
hardy925: `toSerializable` would also have the same issue

strategy/from_json/date.ts
hardy925: We can't fix issues in projects that use this library, and if they are inputting a "string like this" expecting a Date, we can reply on native JS to let them know they messed up.

strategy/from_json/to_object_containing.ts
ChrisDufourMB: Hmm, well if it's at least consistent we can probably just leave it for now.

We may want to look into an overload or something at some point though so we can support constructors with args. This would help when implementors want to gradually move thier classes over to ts_serialize, rather than having to rewrite all their classes without no-args constructors.

hardy925: fixed it for both `toObjectContaining` an `toSerializable`

strategy/from_json/date.ts
hardy925: I have this test passing

```ts

test({
  name: "createDateStrategy - throws on invalid dates",
  fn() {
      @SerializeProperty({
        fromJSONStrategy: createDateStrategy(/^Im going to shoot my foot$/),
      })
      date!: Date;
    }
    try {
      new Test().fromJSON(`{"date":"Im going to shoot my foot"}`);
      fail("Non date string did not error");
    } catch (error) {
      assertEquals(error.message, ERROR_INVALID_DATE);
    }
  },
});
```

strategy/from_json/date_test.ts
```diff
@@ -20,7 +21,25 @@ test({
 });
 
 test({
-  name: "Will not serialize non date strings",
+  name: "createDateStrategy - Will not serialize non date strings",
```
ms1111:
```suggestion
  name: "createDateStrategy - Will not deserialize non date strings",
```

strategy/from_json/date_test.ts
```diff
@@ -20,7 +21,25 @@ test({
...
+      new Test().fromJSON(`{"date":"I am not a date!"}`);
+      fail("Non date string did not error");
+    } catch (error) {
+      assertEquals(error.message, ERROR_INVALID_DATE);
+    }
+  },
+});
+
+test({
+  name: "iso8601Date - Will not serialize non date strings",
```
ms1111:
```suggestion
  name: "iso8601Date - Will not deserialize non date strings",
```

strategy/_utils.ts
```diff
@@ -0,0 +1,26 @@
...
+/** for strategy values */
+export function isObject(obj: any): obj is Record<string, any> {
+  return Object.prototype.toString.call(obj) === "[object Object]";
+}
+
+/** for strategy type arguments */
+function isNewable<T>(
+  type: NewSerializable<T> | FunctionSerializable,
+): type is NewSerializable<T> {
+  return !!type.name;
```
ms1111: Neat.

I did some Googling, what would you think about https://stackoverflow.com/a/57862312/810525
```
function isConstructor(func) {
    return (func && typeof func === "function" && func.prototype && func.prototype.constructor) === func;
}
```
The relevant bit of that is probably `func.prototype && (func.prototype.constructor === func)`.

strategy/from_json/to_object_containing.ts
```diff
@@ -0,0 +1,55 @@
...
+          continue;
+        }
+        if (Array.isArray(value[prop])) {
+          record[prop] = (value[prop] as JSONArray).map((v: JSONValue) => {
+            if (!isObject(v)) {
+              throw new Error(ERROR_TO_OBJECT_CONTAINING_INVALID_SUB_VALUE);
+            }
+            return getNew(type).fromJSON(v);
+          });
+          continue;
```
ms1111: How would you feel about changing the `if / continue / if / continue` to `if / else if ... / else if ... / else`? I watched a talk (from a C++ conference of all things) where someone was explaining why that can be easier to read (makes the control structure obvious just by looking at the indentation, basically).

strategy/to_json/from_oject_containing_test.ts
```diff
@@ -0,0 +1,68 @@
+// Copyright 2018-2020 Gamebridge.ai authors. All rights reserved. MIT license.
```
ms1111: Typo in filename - from_oject -> from_object

strategy/to_json/from_object_containing.ts
```diff
@@ -0,0 +1,16 @@
...
+import { JSONObject, JSONValue, Serializable } from "../../serializable.ts";
+
+export function fromObjectContaining(
+  value: Record<string, Serializable | Serializable[]>,
+): JSONObject {
+  const record: Record<string, JSONValue> = {};
+  for (const prop in value) {
+    record[prop] = Array.isArray(value[prop])
+      ? (value[prop] as Serializable[]).map((v) => v.tsSerialize())
+      : (value[prop] as Serializable).tsSerialize();
```
ms1111, line 13: I thought this already happened by default. Maybe just for arrays? I figure if we have the objects in memory, we know if they are Serializable, so the default strategy should be to serialize _really hard_. Recursively go through each value in an embedded array or object and `tsSerialize()` it.

But again I thought we already did that by default (at least one level deep). Maybe we only did it for arrays?

ms1111 (commented):
I wonder if fromObjectContaining() needs to exist or if we should just always do it. Looks really good other than that.

hardy925:
Added requested changes, also added `fromObjectContaining` as a `toJSONstrategy`,  @ChrisDufourMB your change also has type protection.

Screen Shot 2020-12-10 at 11 59 22 PM

strategy/to_json/from_object_containing.ts
hardy925: we do this by default with toJSON but that returns a string and parsing causes escaping, by using `tsSerialize` we are avoiding that issue.

strategy/from_json/to_object_containing.ts
hardy925: I am a fan of exit early exit often. We're not doing crazy processing so the shortcircuit seems appropriate.

strategy/_utils.ts
```diff
@@ -0,0 +1,27 @@
...
+/** for strategy type arguments */
+function isNewable<T>(
+  type: NewSerializable<T> | FunctionSerializable,
+): type is NewSerializable<T> {
+  return (type && typeof type === "function" && type.prototype &&
+    type.prototype.constructor) === type;
+}
+
+/** get new strategy type arguments */
+export function getNew<T>(
```
ChrisDufourMB, line 23: 💯 , may look into adding this to the polymorphic initializer too

ChrisDufourMB (approved):

PR #92, opened 2020-12-16
fixes #91

### LINKED ISSUE #91: [BUG] - Embedded class fields overwrite containing class fields of the same name
**Describe the bug**
If you have a field `field1` in a class, and it embeds another class with the same field name, when deserializing you get the wrong value in the parent class.

**To Reproduce**
```ts
test({
  name: "Nested fields shouldn't overwrite containing class fields of the same name",
  fn() {
    class Embedded extends Serializable {
      @SerializeProperty()
      public field1?: string;
    }

    class Root extends Serializable {
      @SerializeProperty()
      field1?: string;
      @SerializeProperty({ fromJSONStrategy: toSerializable(Embedded) })
      public embedded?: Embedded[];
    }

    const input = {
      field1: "field_value_in_outer_class",
      embedded: [
        { field1: "field_value_in_inner_class" },
      ]
    }

    const root = new Root().fromJSON(input);
    assertEquals(root.field1, "field_value_in_outer_class"); // i.e. not field_value_in_inner_class
  }
});
```

Fails with:
```diff
    [Diff] Actual / Expected


-   "field_value_in_inner_class"
+   "field_value_in_outer_class"
```

**Expected behaviour**
In deeply nested structures, the same field name may appear multiple times and no overwriting should occur

hardy925:
**Description**

different parsing logic to fix #91 

- new parser 
- expose forgotten interfaces

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/serializable.ts
+++ b/serializable.ts
@@ -179,6 +179,7 @@ function fromJSON<T>(
   const serializablePropertyMap = SERIALIZABLE_CLASS_MAP.get(
     context?.constructor?.prototype,
   );
+
   if (!serializablePropertyMap) {
     throw new Error(
       `${ERROR_MISSING_PROPERTIES_MAP}: ${context?.constructor
@@ -197,10 +198,12 @@ function fromJSON<T>(
     const processedValue: unknown = Array.isArray(value)
       ? value.map((v) => fromJSONStrategy(v))
       : fromJSONStrategy(value as JSONValue);
+
     if (propertyKey) {
       accumulator[propertyKey as keyof T] = processedValue as any;
     }
   }
+
   return Object.assign(
     context,
     accumulator as T,
--- a/serializable.ts
+++ b/serializable.ts
@@ -186,35 +186,23 @@ function fromJSON<T>(
     );
   }
 
-  const _json = typeof json === "string" ? json : JSON.stringify(json);
+  const _json = typeof json === "string" ? JSON.parse(json) : json;
+  const accumulator: Partial<T> = {};
 
+  for (const [key, value] of Object.entries(_json)) {
+    const {
+      propertyKey,
+      fromJSONStrategy = fromJSONDefault,
+    } = serializablePropertyMap.getBySerializedKey(key) || {};
+    const processedValue: unknown = Array.isArray(value)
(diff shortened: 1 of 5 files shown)


ChrisDufourMB (dismissed):

serializable.ts
```diff
@@ -179,42 +179,33 @@ function fromJSON<T>(
...
+  const accumulator: Partial<T> = {};
+
+  for (const [key, value] of Object.entries(_json)) {
+    const {
+      propertyKey,
+      fromJSONStrategy = fromJSONDefault,
+    } = serializablePropertyMap.getBySerializedKey(key) || {};
+    const processedValue: unknown = Array.isArray(value)
+      ? value.map((v) => fromJSONStrategy(v))
+      : fromJSONStrategy(value as JSONValue);
```
ms1111: Should we nest this inside the `if (propertyKey)` below, so we don't process it if it's not a `@SerializeProperty()`?

serializable.ts
```diff
@@ -179,42 +179,33 @@ function fromJSON<T>(
...
+
+    if (propertyKey) {
+      accumulator[propertyKey as keyof T] = processedValue as any;
```
ms1111: If we have `@SerializeProperty()` with no string value (as opposed to `@SerializeProperty("something")`), will this still fire (will `propertyKey` still be defined)?

ms1111 (approved):
🚀

serializable.ts
```diff
@@ -179,42 +179,34 @@ function fromJSON<T>(
...
+    const {
+      propertyKey,
+      fromJSONStrategy = fromJSONDefault,
+    } = serializablePropertyMap.getBySerializedKey(key) || {};
 
-        const {
-          propertyKey,
-          fromJSONStrategy = fromJSONDefault,
-        } = serializablePropertyMap.getBySerializedKey(key) || {};
+    if (!propertyKey) {
```
ms1111, line 199: If we have @SerializeProperty() with no string value (as opposed to @SerializeProperty("something")), will this still be OK (will propertyKey still be defined)?

hardy925: yep! because the key is the property as string if not set

PR #95, opened 2020-12-27
100% test coverage

hardy925:
**Description**

run `deno test --coverage --unstable`

Screen Shot 2020-12-27 at 4 27 58 AM

- 100% test coverage

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/serializable.ts
+++ b/serializable.ts
@@ -168,17 +168,6 @@ function fromJSON<T>(
   context: Serializable,
   json: string | JSONValue | Object,
 ): T {
-  const serializablePropertyMap = SERIALIZABLE_CLASS_MAP.get(
-    context?.constructor?.prototype,
-  );
-
-  if (!serializablePropertyMap) {
-    throw new Error(
-      `${ERROR_MISSING_PROPERTIES_MAP}: ${context?.constructor
-        ?.prototype}`,
-    );
-  }
-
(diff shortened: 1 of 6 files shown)


serializable.ts
```diff
@@ -168,25 +168,16 @@ function fromJSON<T>(
...
     const {
       propertyKey,
       fromJSONStrategy = fromJSONDefault,
-    } = serializablePropertyMap.getBySerializedKey(key) || {};
+    } = (SERIALIZABLE_CLASS_MAP.get(
+      context?.constructor?.prototype,
+    ) as SerializePropertyOptionsMap).getBySerializedKey(key) || {};
 
     if (!propertyKey) {
       continue;
```
ms1111, line 183: I think this change means:
- previously, if you somehow called `fromJSON()` on a non-Serializable constructor (which should be impossible anyway, given compiler restrictions), you'd get an ERROR_MISSING_PROPERTIES_MAP
- now, it will not throw an error but will give you an empty object

Did I read it right? Is that how you want it to behave?

ms1111 (dismissed):

serializable.ts
hardy925: This is a private function exposed via `serializable.fromJSON`, which always calls with `this` as the first argument, basically the error that was thrown was unreachable code. From that, I didn't need to declare the variable so I move the iterator to the for loop.

PR #120, opened 2021-02-08
#118 use serialized key for poly switch

### LINKED ISSUE #118: [BUG] - Property key transforms are not applied when using @PolymorphicSwitch
**Describe the bug** A clear and concise description of what the bug is.

When resolving a polymorphic switch the property key is used, but neither the `tsTransformKey` or any `@SerializeProperty` key transform logic is applied. This matters for `@PolymorphicSwitch` as it targets a property of the class, and not an arbitrary key of the serialized object.

```ts
test({
  name:
    "polymorphic switch should apply key transforms on switched property names based on the target property's serialization logic",
  fn() {
    abstract class AbstractClass extends Serializable {
      public tsTransformKey(key: string): string {
        return "__" + key;
      }
    }

    class TestClass extends AbstractClass {
      public tsTransformKey(key: string): string {
        return "+" + key;
      }
      @PolymorphicSwitch(() => new TestClass())
      public static class = "TestClass";
      @SerializeProperty()
      public someProperty = "original value";
    }

    class TestClass2 extends AbstractClass {
      @SerializeProperty("some_property")
      @PolymorphicSwitch(() => new TestClass(), "some value")
      public someProperty = "some value";
    }

    const testData = { "+class": "TestClass", "+someProperty": "new value" };
    const polyClass = polymorphicClassFromJSON(AbstractClass, testData);

    assert(polyClass instanceof TestClass);
    assertEquals((polyClass as TestClass).someProperty, "new value");

    const testData2 = { "some_property": "some value" };
    const polyClass2 = polymorphicClassFromJSON(AbstractClass, testData2);

    assert(polyClass2 instanceof TestClass2);
    assertEquals((polyClass2 as TestClass2).someProperty, "some value");
  },
});
```
**Expected behaviour** A clear and concise description of what you expected to
happen.

Any relevant key transform logic should be applied when comparing a property key to the key in the serialized object.

ChrisDufourMB:
**Description**

Note: This is a breaking change - `@PolymorphicSwitch` no longer supports static properties

Refactor the polymorphic switch decorator to use the same key serialization logic as the `@SerializeProperty` decorator.

```ts
@SerializeProperty()
@PolymorphicSwitch(() => new TestClass(), "TestClass")
```

Enhance the polymorphic switch decorator to accept a `test` function, that allows for custom logic when determining if the target property means that the polymorphic type is this class.
```ts
@PolymorphicSwitch(
  () => new TestClassOtherYear(),
  (propertyValue) => (propertyValue as Date).getFullYear() !== 2020,
)
```
Fixes #118

Renames `input` argument to `json` for polymorphic decorator functions

**Things to look at**

- [ ] Test coverage
- [ ] Code Style
- [ ] Documentation (`README.md`, `doc/`, `examples/`, etc..)


hardy925 (changes requested):
Hey hey,

I am requesting changes because I ran the build script and our node tests failed.

Screen Shot 2021-02-11 at 5 22 30 PM

We are also missing two tests, if we cannot get a code path to those lines then we can probably make a small update 

Screen Shot 2021-02-11 at 4 59 35 PM

Thanks!

hardy925 (approved):
This looks good to me -  100% test coverage, all passing and our npm build looks good on the latest deno. 

Thanks for the contribution

PR #122, opened 2021-02-12
Force CI to test against ubuntu-20.04

ChrisDufourMB:
**Description**
Force upgrade CI runners to ubuntu-20.04, resolving a build issue seen in #120 
Reformat all the `.yml` files. Style changes were done automagically by VSCode's formatter, so if they are inappropriate I will reverse them.

**Proposed changes in this PR**

- Change ubuntu CI version from `ubuntu-latest` to `ubuntu-20.04`
- Run VSCode's formatter against all the `.yml` files in the project

**Things to look at**

- [ ] Test coverage
- [ ] Code Style
- [ ] Documentation (`README.md`, `doc/`, `examples/`, etc..)

PR #123, opened 2021-02-12
Revert "Force CI to test against ubuntu-20.04"

ChrisDufourMB:
Reverts GameBridgeAI/ts_serialize#122


hardy613:
ouefff messy, but all good, I think yuo need to update the settings again, damn. I'll look into making that better in the future

hardy613 (approved):
lg, once ubuntu is nailed down does it make sense to go back to `latest` over `20.04`

PR #126, opened 2021-05-05
ISSUE-125 Handle arrays with to/fromJSONStrategy properly

### LINKED ISSUE #125: [BUG] - Using strategies on Array types runs the strategy on array elements instead of the whole array
**Describe the bug** 

**To Reproduce**
```
public myProp:{id: number}[];
...

}

```

**Expected behaviour** 

khan-saim:
**Description**

Issue fixed: [125](https://github.com/GameBridgeAI/ts_serialize/issues/125)

Here we change the to/fromJSONStrategy handling when the property is an array. If a strategy is provided, it is applied on the full array rather than individually on the elements of the array. 

**Things to look at**

- [ ] Test coverage
- [ ] Code Style
- [ ] Documentation (`README.md`, `doc/`, `examples/`, etc..)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ examples/node/*
 coverage
 coverage.lcov
 .DS_Store
+.vscode
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,1 +1,6 @@
-{ "deno.enable": true }
+{
+    "deno.enable": true,
+    "deno.lint": true,
+    "deno.unstable": false,
+    "deno.path": "/Users/saim.khan/.deno/bin/deno"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-    "deno.enable": true,
-    "deno.lint": true,
-    "deno.unstable": false,
-    "deno.path": "/Users/saim.khan/.deno/bin/deno"
-}
--- a/serializable.ts
+++ b/serializable.ts
@@ -152,12 +152,17 @@ export function toPojo(
     }
 
     if (Array.isArray(value)) {
+      let notSerializableElements = false;
       record[serializedKey] = value.map((item: any) => {
         if (item instanceof Serializable) {
           return toPojo(item);
         }
-        return toJSONStrategy(item);
+        notSerializableElements = true;
+        return item;
       });
+      if(notSerializableElements) {
+        record[serializedKey] = toJSONStrategy(value);
+      }
     } else if (value !== undefined) {
       record[serializedKey] = toJSONStrategy(value);
     }
@@ -181,7 +186,7 @@ function fromJSON<T>(
   for (const [key, value] of Object.entries(_json)) {
     const {
       propertyKey,
-      fromJSONStrategy = fromJSONDefault,
+      fromJSONStrategy,
     } = SERIALIZABLE_CLASS_MAP.get(
       context?.constructor?.prototype,
     )?.getBySerializedKey(key) || {};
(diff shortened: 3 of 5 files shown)


.vscode/settings.json
```diff
@@ -1 +1,6 @@
-{ "deno.enable": true }
+{
+    "deno.enable": true,
+    "deno.lint": true,
+    "deno.unstable": false,
+    "deno.path": "/Users/saim.khan/.deno/bin/deno"
+}
```
khan-saim: oh wait what, why did this sneak in 😭

.gitignore
```diff
@@ -7,4 +7,5 @@ examples/node/*
 !examples/node/README.md
 coverage
 coverage.lcov
-.DS_Store
\ No newline at end of file
+.DS_Store
+.vscode
```
hardy925, line 11: We want to keep  `.vscode`, what issues are you seeing?

.vscode/settings.json
```diff
@@ -1 +0,0 @@
-{ "deno.enable": true }
```
hardy925, line 1: we want this guy too

serializable.ts
```diff
@@ -152,12 +152,19 @@ export function toPojo(
     }
 
     if (Array.isArray(value)) {
+      let notSerializableElements = false;
       record[serializedKey] = value.map((item: any) => {
         if (item instanceof Serializable) {
           return toPojo(item);
         }
```
hardy925, line 159: so. I don't think any of this stuff is really needed now

serializable.ts
```diff
@@ -152,12 +152,19 @@ export function toPojo(
...
-        return toJSONStrategy(item);
+
+        // if item is not Serializable we will apply the toJSONStrategy on the whole array so return items as is
+        notSerializableElements = true;
```
hardy925, line 162: not sure if we need this guy too

serializable.ts
```diff
@@ -152,12 +152,19 @@ export function toPojo(
...
+        return item;
       });
+      if(notSerializableElements) {
+        record[serializedKey] = toJSONStrategy(value);
+      }
```
hardy925, line 167: this can be simplified now, and I don't think we need the check

serializable.ts
```diff
@@ -181,7 +188,7 @@ function fromJSON<T>(
   for (const [key, value] of Object.entries(_json)) {
     const {
       propertyKey,
-      fromJSONStrategy = fromJSONDefault,
+      fromJSONStrategy,
```
hardy925, line 191: I think we'll want to revert this change

serializable.ts
```diff
@@ -190,9 +197,16 @@ function fromJSON<T>(
...
-    accumulator[propertyKey as keyof T] = Array.isArray(value)
-      ? value.map((v) => fromJSONStrategy(v))
-      : fromJSONStrategy(value as JSONValue);
+    // if strategy is not provided we apply the default on the value/elements of the array
+    if (!fromJSONStrategy) {
+      const defaultFromJSONStrategy = fromJSONDefault;
+      accumulator[propertyKey as keyof T] = Array.isArray(value)
+      ? value.map((v) => defaultFromJSONStrategy(v)) : defaultFromJSONStrategy(value as JSONValue)
+      continue;
+    }
```
hardy925, line 206: this can most likely be removed now too

serializable.ts
```diff
@@ -190,9 +197,16 @@ function fromJSON<T>(
...
+
+    // if a strategy is provided we apply the strategy on the entire array
+    accumulator[propertyKey as keyof T] = fromJSONStrategy(value as JSONValue);
```
hardy925, line 209: this is basically it now because we want to pass the direct value through the strat array or not, and we and guarantee the default set by the destructuring.

serialize_property_test.ts
```diff
@@ -560,3 +560,14 @@ test({
...
+test({
+  name: "custom toJSONStrategy with an array of data",
+  fn() {
+    class Test extends Serializable {
+      @SerializeProperty({ toJSONStrategy: (v) => v.join(".") })
+      public test_property = [0,1];
+    }
+    assertEquals(new Test().toJSON(), `{"test_property":"0.1"}`);
+  },
+});
```
hardy925, line 573: this is a good test, but I think we can also remove some other tests

strategy/from_json/to_serializable_test.ts
```diff
@@ -104,7 +104,7 @@ test({
     }
     class Test extends Serializable {
       @SerializeProperty({
-        fromJSONStrategy: (v) => new OtherClass().fromJSON(v),
+        fromJSONStrategy: (v) => (v as JSONArray).map(el => new OtherClass().fromJSON(el)),
```
hardy925, line 107: so you can type this at the param level and it should apply to the function scope

hardy925 (changes requested):
Hey @khan-saim :wave:

Thanks for opening a PR, I did a quick skim of the code and I think we still have some changes to make. For the most part, I believe we would be removing more code than adding.

Let me know if you have questions

Cheers,

shardyMBAI:
Hey @khan-saim, Thanks again for the PR, but I am going to close it as we've recently pushed some rather larger changes, while doing these changes I also took care of this issue with #131

hardy925:
Hey @khan-saim, Thanks again for the PR, but I am going to close it as we've recently pushed some rather larger changes, while doing these changes I also took care of this issue with #131

PR #128, opened 2021-06-05
[iss-127] remove array as an option from strategies

### LINKED ISSUE #127: [Change] - Remove option to pass an array as a strategy
Currently, authors can pass an array as a strategy option, not only does this clutter our code but with the option to use `composeStrategy` this would not be a larger change on authors already using the library. 

The change would have the `tsc` fail on these tests:

```ts

test({
  name: "fromJSONStrategy with array",
  fn() {
      @SerializeProperty({ fromJSONStrategy: [(v) => v] })
      public test_property = 0;
    }
    assertEquals(new Test().fromJSON({ test_property: 99 }).test_property, 99);
  },
});

test({
  name: "toJSONStrategy with array",
  fn() {
      @SerializeProperty({ toJSONStrategy: [(v) => v] })
      public test_property = 0;
    }
    assertEquals(new Test().toJSON(), `{"test_property":0}`);
  },
});
```
as the array would not be an option. THe tests will be removed with the change and authors can use `composeSrategy` in place of an array.

hardy925:
**Description**
Mostly cleaning up some code 

- fixes #127 

- removes array as a strategy, use `composeStrategy` instead
- `fmt` with deno@1.10.3
- turn off unstable code coverage
- remove `JSONArray` for `JSONValue[]`

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
+++ b/.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
@@ -2,9 +2,6 @@ declare module "@gamebridgeai/ts_serialize" {
   /** A JSON object where each property value is a simple JSON value. */
   export type JSONObject = { [key: string]: JSONValue };
 
-  /** A JSON array where each value is a simple JSON value. */
-  export interface JSONArray extends Array<JSONValue> {}
-
   /** A property value in a JSON object. */
   export type JSONValue =
     | string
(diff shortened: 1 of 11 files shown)


ms1111 (approved):
Makes sense to me; I don't think we used the array option.

PR #131, opened 2021-06-17
pass whole array to strategies and add clone

### LINKED ISSUE #125: [BUG] - Using strategies on Array types runs the strategy on array elements instead of the whole array

### LINKED ISSUE #132: Add clone functionality
Allow authors to clone an object, should return a new object deep copied

hardy925:
**Description**

pass arrays to strategies, fixes #125, fixes #132

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/serializable.ts
+++ b/serializable.ts
@@ -40,6 +40,11 @@ export declare interface Serialize {
   tsSerialize(): JSONObject;
 }
 
+/** deep copy `this` */
+export declare interface Clone {
+  clone(jsonObject: Partial<this>): this;
+}
+
 /** Recursively set default serializer logic for own class definition and parent definitions if none exists */
 function getOrInitializeDefaultSerializerLogicForParents(
   targetPrototype: Serializable,
@@ -87,7 +92,6 @@ function getOrInitializeDefaultSerializerLogicForParents(
  *       example.tsSerialize()
  */
 export abstract class Serializable {
-  /** allow empty class serialization */
   constructor() {
     getOrInitializeDefaultSerializerLogicForParents(this.constructor.prototype);
   }
@@ -102,6 +106,14 @@ export abstract class Serializable {
   }
   public tsSerialize(): JSONObject {
     return toPojo(this);
+  }
+  public clone(jsonObject: Partial<this> = {}): this {
+    return Object.assign(
+      new (Object.getPrototypeOf(this)).constructor().fromJSON(
+        this.tsSerialize(),
+      ),
+      jsonObject,
+    );
   }
 }
 
--- a/serializable.ts
+++ b/serializable.ts
@@ -108,8 +108,9 @@ export abstract class Serializable {
     return toPojo(this);
   }
   public clone(jsonObject: Partial<this> = {}): this {
+    const copy = Object.getPrototypeOf(this).constructor;
     return Object.assign(
-      new (Object.getPrototypeOf(this)).constructor().fromJSON(
+      new copy().fromJSON(
         this.tsSerialize(),
       ),
       jsonObject,
--- a/serializable.ts
+++ b/serializable.ts
@@ -148,16 +148,7 @@ export function toPojo(
       toJSONStrategy = toJSONRecursive;
     }
 
-    if (Array.isArray(value)) {
-      record[serializedKey] = value.map((item: JSONValue) => {
-        if (item instanceof Serializable) {
-          return toPojo(item);
-        }
-        return toJSONStrategy(item);
-      });
-    } else if (value !== undefined) {
-      record[serializedKey] = toJSONStrategy(value);
-    }
+    record[serializedKey] = toJSONStrategy(value);
   }
   return record;
 }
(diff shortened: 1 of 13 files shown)


strategy/to_json/from_serializable_test.ts
```diff
@@ -0,0 +1,71 @@
...
+      serializeMe = 999;
+    }
+    class Test2 extends Serializable {
+      @FromSerializable("outer_property")
+      nested: Test1[] = [new Test1()];
+    }
+
+    class Test3 extends Serializable {
+      @FromSerializable("outer_outer_property")
+      nested2: Test2[] = [new Test2()];
```
ms1111, line 37: Is the `@FromSerializable` needed here? I would think if we had
```
@SerializeProperty("outer_outer_property")
nested2: Test2[] = [new Test2()];
```
... then the default strategy would map through the array and call tsSerialize() if it's serializable.

serializable.ts
```diff
@@ -148,16 +148,7 @@ export function toPojo(
...
-      record[serializedKey] = value.map((item: JSONValue) => {
-        if (item instanceof Serializable) {
-          return toPojo(item);
-        }
-        return toJSONStrategy(item);
-      });
-    } else if (value !== undefined) {
-      record[serializedKey] = toJSONStrategy(value);
-    }
+    record[serializedKey] = toJSONStrategy(value);
```
ms1111: I think there's an important change to the default behaviour for serialization here. Previously, if you had an array property and were using the default strategy:

```
@SerializeProperty()
private someArray: SomeClass[];
```
... we'd automatically call toPojo() on each element in the array.

Now the default is that we'll call toJSONStrategy(), which just returns the original value. So serialization that would have worked properly before by default, no longer will.

Granted, you always had to give us more information for _deserialization_ to work, using `toSerializable()` in a `fromJSONStrategy`. But previously you could get away without a `toJSONStrategy` and it would do the right thing, kind of by accident, if you only cared about that direction. So that might be worth updating in the examples.

For our own use, we use a shortcut decorator `@DeserializeAs()`:

```ts
export function DeserializeAs<T>(type: SerializableConstructor<T>, propertyName?: string): PropertyDecorator {
    const objectSerializer: SerializePropertyArgument = {
        fromJSONStrategy: toSerializable(type),
    };
    if (propertyName) {
        objectSerializer.serializedKey = propertyName;
    }

    return SerializeProperty(objectSerializer);
}
```
... we would just need to change it to specify a toJsonStrategy of fromSerializable(), so it's not a big change.

ms1111 (dismissed):

strategy/to_json/from_serializable_test.ts
hardy925: ah so this change runs against the whole array now and not per item, because under the hood `toJSON()` is called and returns the string, so we need to manually say to use `tsSerialize`. This is actually done on purpose because JSON is a text format, and we are trying hard to follow JS API's. for example 

```ts
new Date().toJSON() // returns string value and not a POJO
```

serializable.ts
hardy925: Yes 💯 

This is a breaking change and will not be released until v2 is ready. There are a few other features for us to take care of and I only recently and started to set up and plan v2. I am going to add `new Example().clone(arg: Partial<Example>)` or something similar, and we are going to start supporting common decorators, although I'd like to discuss how that will look before starting.

hardy925: Another breaking change is with `.fromJSON()` it only takes a `JSONValue` now, IDK if this is the best approach but I currently believe it is better than using `Object` to support angular and rather have integrators use typing with angular.

Again that is also up for discussion.

hardy925: oh damn! I just noticed I am bringing in a bug by allowing `undefined` to be processed by strategies. 

Thanks, @ms1111

serializable.ts
```diff
@@ -103,6 +107,15 @@ export abstract class Serializable {
   public tsSerialize(): JSONObject {
     return toPojo(this);
   }
+  public clone(jsonObject: Partial<this> = {}): this {
```
ChrisDufourMB, line 112: I feel that we need a JSDoc comment here explaining what the argument does. It may not be obvious to users who use ts_serialize differently than we do.

serializable.ts
```diff
@@ -40,6 +40,11 @@ export declare interface Serialize {
   tsSerialize(): JSONObject;
 }
 
+/** deep copy `this` */
+export declare interface Clone {
+  clone(jsonObject: Partial<this>): this;
+}
+
```
ChrisDufourMB: Is this interface being used anywhere? Is this different from the method on `Serializable`?

ChrisDufourMB (dismissed):
Changes LGTM, have a few comments surrounding the clone functionality

serializable.ts
hardy925: I made these interfaces for implementing if authors wanted to, thinking about it now though I am unsure if we need clone...

serializable.ts
hardy925: added on the interrface

serializable.ts
hardy925: keeping for now, we can chat about it before v2 release

PR #133, opened 2021-11-12
up deno to latest, test builds, fmt change

hardy925:
**Description**

Doing some chore work, also adding in some changes I've wanted to make for v2. 

- deno@1.16.1
- exported strategies all return a function
- moved node_tests folder, removed examples as we don't really have any

**Things to look at**

- [x] Test coverage
- [x] Code Style


strategy/from_json/date.ts
```diff
@@ -5,7 +5,7 @@ import { ERROR_INVALID_DATE } from "../../error_messages.ts";
 
 /** allows authors to pass a regex to parse as a date */
 export function createDateStrategy(regex: RegExp): FromJSONStrategy {
-  return function _createDateStrategy(input: JSONValue): Date {
+  return (input: JSONValue): Date => {
```
ms1111, line 8: I'm trying to remember if there was a benefit to having it as a named function before. Maybe how it would show up in stack traces?

.github/workflows/scripts/npm_release_files/ts_serialize.d.ts
```diff
@@ -102,27 +102,23 @@ declare module "@gamebridgeai/ts_serialize" {
   ): FromJSONStrategy;
 
   /** serialize data using `tsSerialize` on a subclass Serializable type */
-  export function fromSerializable(
-    value: Serializable | Serializable[],
-  ): JSONValue;
+  export function fromSerializable(): ToJSONStrategy;
```
ms1111, line 105: This does seem cleaner

.github/workflows/ts_serialize_tests.yml
```diff
@@ -39,24 +39,21 @@ jobs:
...
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-        # if: matrix.os == 'ubuntu-latest'
-        if: false
+      # - name: Report code coverage
+      #   uses: zgosalvez/github-actions-report-lcov@v1
+      #   with:
+      #     coverage-files: ts_serialize_coverage.lcov
+      #     minimum-coverage: 90
+      #     artifact-name: code-coverage-report
+      #     github-token: ${{ secrets.GITHUB_TOKEN }}
```
ms1111: Some commented out stuff here

CHANGELOG.md
```diff
@@ -6,23 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.3.2] - unreleased
 
+### Changed
+
+- public strategies are all functions to be run to get the runtime function
+  `iso8601Date` => `iso8601Date()`
```
ms1111, line 15: Is this just for consistency, since before some of them were functions and some of them returned functions?

strategy/to_json/from_object_containing.ts
```diff
@@ -1,17 +1,20 @@
...
-      ? (value[prop] as Serializable[]).map((v) => v.tsSerialize())
-      : (value[prop] as Serializable).tsSerialize();
-  }
+/** convert `{ [_: string]: Serializable }` to `{ [_: string]: Serializable.tsSerialize() }` */
+export function fromObjectContaining(): ToJSONStrategy {
+  return (
+    value: Record<string, Serializable | Serializable[]>,
+  ): JSONObject => {
+    const record: Record<string, JSONValue> = {};
+    for (const [prop, obj] of Object.entries(value)) {
```
ms1111, line 12: Nice yeah, we don't need to support IE11 anymore

image

ms1111 (dismissed):

strategy/to_json/from_object_containing.ts
hardy925: Yea, it also allows the compiler to infer `obj`'s type, instead of us having to cast it with `value[prop]`. so its nice that way as well

strategy/from_json/date.ts
hardy925: I think that is correct, this way `anon` function shows up. `createDateStrategy` will also be in the stack as the parent caller. It seems like the right time to use an arrow function after thinking about the use case. Thoughts?

.github/workflows/ts_serialize_tests.yml
hardy925: nice, thanks

CHANGELOG.md
hardy925: Exactly yes.

strategy/from_json/date.ts
ms1111: Yeah, that’s true, the parent function in the call trace will show up. So that seems fine.

ChrisDufourMB (approved):
LGTM

PR #137, opened 2022-03-16
[v2] release prep

### LINKED ISSUE #121: [BUG] - run `deno lint --unstable`
## Description

Deno lint is available (though `--unstable` is needed) and when we run the command our code base fails 🤷 

### Steps to reproduce the behaviour

From the root

- `rm -rf examples/node/node_modules examples/node/node.js dist/ deno.d.ts`
- `deno lint --unstable` 
- observe failures

### A clear and concise description of the expected behaviour
- `rm -rf examples/node/node_modules examples/node/node.js dist/`
- `deno lint --unstable` 
- observe the passing command

### Desktop 
- all environments

### Additional context

Although `--unstable` I think we can make use of this command like we do `deno test --coverage --unstable`

### LINKED ISSUE #125: [BUG] - Using strategies on Array types runs the strategy on array elements instead of the whole array

### LINKED ISSUE #132: Add clone functionality

### LINKED ISSUE #134: [BUG] - Clone
- [x] Export clone implementation
- [x] Add clone to docs
- [x] Move all docs to README.md and remove `docs/`

### LINKED ISSUE #135: Convert all sub functions to arrow functions
Currently we still have functions that return `function` rather than `() => {}`. We can convert them to be arrow functions.

https://github.com/GameBridgeAI/ts_serialize/blob/develop/strategy/compose_strategy.ts#L21 is an example of a function we can convert.

hardy613:
**Description**
Preparation for v2 release.

For reference the full changelog for v2 is:

- _breaking_ remove JSONArray for JSONValue[]
- _breaking_ fromJSON only accepts `JSONValue` 
- _breaking_ fix #125 strategies are now passed the full array, this allows more
  flexibility for conversions
- _breaking_ removes array as a strategy, use `composeStrategy` instead
- move `docs/` to `README.md` and delete folder
- fixes #135
- fixes #134
- std@0.129.0
- public strategies are all functions to be run to get the runtime function
  `iso8601Date` => `iso8601Date()`
- deno@1.19.x
- fmt with deno@1.19.x
- turn off unstable code coverage
- add type guard in `getNewSerializable` fixing compile issue
- fix #121 address lint issues
- add `fromSerializable` strategy
- fix #132 adding `serializable.clone(json: JSONObject): this`

Once merged we'll need to update the changelog once more with the release date

**Things to look at**

- [x] Test coverage
- [x] Code Style


strategy/compose_strategy.ts
```diff
@@ -18,10 +18,8 @@ export function composeStrategy(
     | FromJSONStrategy
   )[]
 ): FromJSONStrategy | ToJSONStrategy {
-  return function _composeStrategy(
```
ChrisDufourMB, line 21: I think the original reason we did this was to that in exceptions we would get a named function to show up. I have no strong opinion one way or the other though.

.gitignore
```diff
@@ -1,11 +1,11 @@
 deno.d.ts
 dist
-node_tests/*
-!node_tests/node_tests.ts
-!node_tests/tsconfig.json
-!node_tests/package.json
-!node_tests/README.md
-!node_tests/fixtures
+.github/node_tests/*
```
ChrisDufourMB, line 3: What's the reasoning behind this change? In addition to the CI aspect I thought that the node tests were also a code example of how you can use ts_serialize in a typescript context, so it made sense to have them as part of the docs.

On the other hand node usage is covered in the README, so I'm fine with this change

README.md
```diff
@@ -19,13 +19,645 @@ A zero dependency library for serializing data.
...
 - [`JSON`](https://www.json.org/json-en.html)
 
-## [Usage](https://gamebridgeai.github.io/ts_serialize/)
+## Usage
 
-- [Installing](https://gamebridgeai.github.io/ts_serialize/installing)
-- [Serializable and SerializeProperty](https://gamebridgeai.github.io/ts_serialize/serializable)
-- [Strategies](https://gamebridgeai.github.io/ts_serialize/strategies)
-- [Global transformKey](https://gamebridgeai.github.io/ts_serialize/transforming)
-- [Polymorphism](https://gamebridgeai.github.io/ts_serialize/polymorphism)
```
ChrisDufourMB, line 28: Fun fact, even if we don't include a table of contents ourselves one is still generated by github we can use:
Screen Shot 2022-03-16 at 15 12 21

ChrisDufourMB (approved):
LGTM!

No blockers, just a few comments.

README.md
hardy613: 🔥

.gitignore
hardy613: basically from conversations, it was trying to be both examples and tests, but not doing a good job of either. This move is to remove it from the root so the docs/README is used. With the changes to the readme now this is only used on `github` so it made sense to move the folder. 

TBH, [the deno -> node conversation](https://github.com/denoland/deno/discussions/9569) has probably moved forward from when we last looked and there might exist a solution that allows us to run the deno tests within a node env.

ChrisDufourMB: Sounds good, I will create a discussion to see if we can find a better way to transpiling into node

strategy/compose_strategy.ts
hardy613: yeap. @ms1111 mentioned this as well and initially, that was the idea, however, the parent function will display in the stack trace which will lead developers to the right location. There are other benefits as well, we are not creating a new object prototype, which also means it cannot be `new`ed and a new function scope is not created.

I think it makes sense that we go with: For global scope, we can use `function`, for `new` constructors we can use `class` and for everything else we can use `() => {}` .

PR #142, opened 2022-03-18
[feature(docs)] Test example code

### LINKED ISSUE #141: [enhancement] - update readme polymorphic docs to use asserts
**Describe the bug**
Currently our docs for polymorphic usage contain code examples that use `console.log`, we should convert these to use asserts from the Deno std. This will help us with #139 .

1. [Go to https://github.com/GameBridgeAI/ts_serialize/](https://github.com/GameBridgeAI/ts_serialize#polymorphism)
2. Scroll down to code examples, 
3. Observe `console.log` usage

**Expected behaviour** 

With #139 we'll be testing our doc's code examples and we'll need asserts.

hardy613:
**Description**
fixes #141 and adds support for #139 - I think I found some bugs in our docs as well while doing this

running `$ deno test --doc` should run all code and test examples

- code in our readme compiles
- code in our readme uses asserts
- fromJSON and ToJSON strategies are removed for a Strategy type 
 

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/.github/workflows/ts_serialize_tests.yml
+++ b/.github/workflows/ts_serialize_tests.yml
@@ -40,7 +40,7 @@ jobs:
         run: deno cache test_deps.ts
 
       - name: Deno test
-        run: deno test
+        run: deno test --doc
 
       - name: Run package.json generation tests
         run: .github/workflows/scripts/npm_release_files/create_npm_package_file_test.sh
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ parameter, the `key` as string and should return a string.
 
 - `serializedKey` (Optional) `{string | ToSerializedKeyStrategy}` - Used as the
   key in the serialized object
-- `toJSONStrategy` (Optional) `{ToJSONStrategy}` - Used when serializing
-- `fromJSONStrategy` (Optional) `{FromJSONStrategy}` - Used when deserializing
+- `toJSONStrategy` (Optional) `{Strategy}` - Used when serializing
+- `fromJSONStrategy` (Optional) `{Strategy}` - Used when deserializing
 
 ```ts
 import { Serializable, SerializeProperty } from "./mod.ts";
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 [![tests](https://github.com/GameBridgeAI/ts_serialize/workflows/tests/badge.svg)](https://github.com/GameBridgeAI/ts_serialize/workflows/tests/badge.svg)
 [![release](https://github.com/GameBridgeAI/ts_serialize/workflows/release/badge.svg)](https://github.com/GameBridgeAI/ts_serialize/workflows/release/badge.svg)
-[![github doc](https://img.shields.io/badge/github-doc-5279AA.svg)](https://gamebridgeai.github.io/ts_serialize)
 [![deno doc](https://doc.deno.land/badge.svg)](https://doc.deno.land/https/deno.land/x/ts_serialize/mod.ts)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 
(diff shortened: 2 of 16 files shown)


.github/workflows/ts_serialize_tests.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         run: deno cache test_deps.ts
 
       - name: Deno test
-        run: deno test
+        run: deno test --doc
```
ChrisDufourMB, line 43: Note that this doesn't actually run the tests in the READMEs, it only does type checking. I wrote a quick script to extract the tests and run them:
`deno run --allow-read --allow-write --allow-run run-doc-tests.ts README.md`
`run-doc-tests.ts`
```ts
import { basename } from "https://deno.land/std@0.127.0/path/mod.ts";

const args = Deno.args;

const testPrefix = [
  'import { test } from "./test_deps.ts";',
  "test({",
  "name:'test',",
  "fn() {",
];

const testPostfix = [
  "},",
  "});",
  "",
];

const docTestsPath = ".";

for (const arg of args) {
  await extractTests(arg);
}

async function extractTests(file: string) {
  const text = await Deno.readTextFile("./README.md");
  const textLines = text.split("\n");

  let testCounter = 0;
  let testLines: string[] | null = null;

  for (const line of textLines) {
    // Start of test
    if (testLines == null && line == "```ts") {
      testLines = [];
    } // End of test
    else if (testLines != null && line == "```") {
      // Create test file
      const outputPath = `${docTestsPath}/${
        basename(file)
      }.${testCounter}.test.ts`;

      const testImportLines = testLines.filter((line) =>
        line.startsWith("import")
      );
      const testBodyLines = testLines.filter((line) =>
        !line.startsWith("import")
      );
      const outputLines = testImportLines.concat(testPrefix).concat(
        testBodyLines,
      ).concat(testPostfix);
      const outputText = outputLines.join("\n");

      console.log(`Writing test to ${outputPath}`);

      await Deno.writeTextFile(outputPath, outputText);

      console.log(`Running tests...`);
      const process = Deno.run({ cmd: ["deno", "test", outputPath] });
      await process.status();

      console.log(`Removing tests...`);
      await Deno.remove(outputPath);

      testLines = null;
      testCounter++;
    } // Body of test
    else if (testLines != null) {
      testLines.push(line);
    }
  }
}
```

ChrisDufourMB: I don't think the intent behind this PR was to run the READMEs as tests, so I think this level of functionality out of scope

strategy/compose_strategy.ts
```diff
@@ -1,25 +1,16 @@
...
-import { JSONValue } from "../serializable.ts";
 
 /** Functions used when hydrating data */
 // deno-lint-ignore no-explicit-any
-export type FromJSONStrategy = (value: JSONValue) => any;
-
-/** Functions used when dehydrating data */
-// deno-lint-ignore no-explicit-any
-export type ToJSONStrategy = (value: any) => JSONValue;
+export type Strategy = (value: any) => any;
```
ChrisDufourMB: What's the reasoning behind this change? With how the previous strategies implementation worked we were able to enforce that input and output be formatted a certain way that we know will be consistent and usable JSON, with this new way the types are less restricted.

For example, now a user can create a toJson strategy where the output is something not serializable, like `Symbol()` or a `function`, resulting in what was previously a TS compile error now being presented as valid.

ChrisDufourMB (commented):
Changes look good to me, but I am a bit concerned about the reduction of type safety as a result of the `Strategy` change

.github/workflows/ts_serialize_tests.yml
hardy925: I made an update to add in the `test()` function, so I think you can reduce this script quite a bit now, thanks man!

README.md
```diff
@@ -101,33 +87,41 @@ parameter, the `key` as string and should return a string.
 
 - `serializedKey` (Optional) `{string | ToSerializedKeyStrategy}` - Used as the
   key in the serialized object
-- `toJSONStrategy` (Optional) `{ToJSONStrategy}` - Used when serializing
-- `fromJSONStrategy` (Optional) `{FromJSONStrategy}` - Used when deserializing
+- `toJSONStrategy` (Optional) `{Strategy}` - Used when serializing
+- `fromJSONStrategy` (Optional) `{Strategy}` - Used when deserializing
```
ChrisDufourMB: I think this needs to be reverted back to `ToJSONStrategy` and `FromJSONStrategy`

README.md
```diff
@@ -137,20 +131,28 @@ any value used for that key will be overridden by the child value. _With
...
 
 ```ts
-class Test1 extends Serializable {
-  @SerializeProperty("serialize_me")
-  serializeMe = "nice1";
-}
+import { Serializable, SerializeProperty } from "./mod.ts";
+import { assertEquals, test } from "./test_deps.ts";
+
+test({
```
ChrisDufourMB, line 137: I think it may be a bit more readable to not use tests in the readme. I did some testing with deno assertions and I think that rather than relying on `deno test` I now think it may be better to just run the sample code and check the processes response code.

I will make a discussion/issue for running the example code,  I think going so far as to rewrite all the example code is out of scope for this ticket. If you want to keep the `test` blocks though I think that's fine for now too.

https://github.com/GameBridgeAI/ts_serialize/discussions/139

ChrisDufourMB (commented):
A few more comments

README.md
hardy613: no hard thoughts on it here, want to keep it for now and we can decide when we add the other script to run the code blocks?

ChrisDufourMB: That's fine with me

ChrisDufourMB (approved):

PR #144, opened 2022-04-03
Test markdown code blocks

hardy613:
**Description**
fixes #139 

We can now run `deno run --allow-read --allow-write --allow-run .github/test_markdown.ts`

- adds a script to: 
  - collect all `.md` files recursively
  - read each `.md` file for typescript code blocks
  - runs each code block as it's own file
- should we only run `test()` code blocks?

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/.github/test_markdown.ts
+++ b/.github/test_markdown.ts
@@ -6,7 +6,7 @@
 
 import { walk } from "https://deno.land/std@0.133.0/fs/mod.ts";
 
-enum CodeBlockTokens {
+enum CodeBlockToken {
   StartTs = "```ts",
   StartTypescript = "```typescript",
   End = "```",
@@ -26,15 +26,15 @@ for await (const { path } of walk(Deno.args[0] || ".", { exts: [".md"] })) {
     currentLine += 1;
     if (
       !readingCodeBlock &&
-      (line === CodeBlockTokens.StartTs ||
-        line === CodeBlockTokens.StartTypescript)
+      (line === CodeBlockToken.StartTs ||
+        line === CodeBlockToken.StartTypescript)
     ) {
       readingCodeBlock = true;
       startLine = currentLine;
       continue;
     }
 
-    if (readingCodeBlock && line === CodeBlockTokens.End) {
+    if (readingCodeBlock && line === CodeBlockToken.End) {
       readingCodeBlock = false;
       testSuite.set(
         path,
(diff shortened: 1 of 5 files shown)


.github/test_markdown.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/** run the testSuites and stop process with `code` when `success` fails */
+for (const [file, tests] of testSuites.entries()) {
+  for (const { startLine, endLine, testCode } of tests) {
+    const path = `${file}:${startLine}-${endLine}.ts`;
+    await Deno.writeTextFile(path, testCode);
+    const { success, code } = await Deno.run({ cmd: ["deno", "test", path] })
+      .status();
+    await Deno.remove(path);
```
ChrisDufourMB: We may want to consider adding a try/catch around the earlier code in case something doesn't go as expected when writing/running the tests.

.github/test_markdown.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  if ((testSuites.get(path) ?? []).length === 0) {
+    testSuites.delete(path);
+  }
+}
+
+/** run the testSuites and stop process with `code` when `success` fails */
+for (const [file, tests] of testSuites.entries()) {
+  for (const { startLine, endLine, testCode } of tests) {
+    const path = `${file}:${startLine}-${endLine}.ts`;
```
ChrisDufourMB, line 82: Have we considered writing to a temp dir instead of in-place? I can foresee a situation where a dev may start running the tests, then commit their code, which could then also commit these temporary files.

Another option may be passing them to test via their STDIN (I tried this and couldn't get it to work), or possibly some unix IPC socket magic.

ChrisDufourMB (dismissed):
Tested it and it works great 💯 

I left some comments about some possible improvements, mainly around writing and reading the generated test files.

.github/test_markdown.ts
hardy613: I tried this with `deno.makeTempDir`, the problem is we then need to start messing with import path logic and its just wildly out of scope IMHO. STDIN would have been great but `deno.test` only supports files or directory locations, no streams :unamused:

.github/test_markdown.ts
hardy613: not catching, but using a `finally` block to remove the file and run the process check.

ChrisDufourMB (approved):

PR #145, opened 2022-04-05
Improve transpiling to node

### LINKED ISSUE #143: Improve transpiling from deno to node
### Discussed in https://github.com/GameBridgeAI/ts_serialize/discussions/138

<div type='discussions-op-text'>

<sup>Originally posted by **ChrisDufourMB** March 16, 2022</sup>
Many deno projects are interested in publishing node compatible code, including us. Our current solution involves transpiling the result of `deno bundle` into something node can use using babel. While this works, it's potentially unstable, as changes to either deno or babel could result in unexpected changes.

We should look into more stable deno to node traspilers as they would likely have the consistent support needed to handle changes to deno going forward.

One example is `dnt` (Deno to Node) by the Deno developers:
https://github.com/denoland/dnt</div>

ChrisDufourMB:
**Description**
Refactor the transpiling to node process to use `dnt` instead of our custom babel setup. This is a big improvement as the previous setup relied on some interesting tricks to get everything to work, including downloading a traspiled version of babel. This new setup is a simple deno dependancy maintained by the deno developers themselves.

Fixes #143

- Remove previous transpiling setup
- Adds a new transpiling setup using [dnt](https://github.com/denoland/dnt)
- Remove node tests, dnt transpiles and runs all the deno tests in node
- Update CI to use new scripts
- Update deno std packages to the latest version
- Some minor code cleanup
- Update node release build to use node 16

To compile to node manually run `deno run -A .github/workflows/scripts/build_npm.ts -v "v0.0.0-test"`, then check `.github/workflows/dist`

**Things to look at**

- [ ] Test coverage
- [ ] Code Style


.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
...
+  `\n\t-o,  --out=[path]         The path to the target directory where the built files should be placed. Defaults to "${outDirDefault}"`;
+
+function printHelpText() {
+  console.log(helpText);
+}
+
+async function main() {
+  const flags = parse(Deno.args);
+
+  if (flags["h"] || flags["help"]) {
```
hardy613: stylistic nit: dot-notation? not a blocker as it passes lint, but it is a preference

.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
...
+const outDirDefault = "./.github/workflows/dist";
+
+const helpText =
+  "Deno to node transpiler, converts the ts_serialize module to a node compatible module, and writes it to .github/workflows/npm, or the directory of your choice." +
+  "\n" +
+  "\nThis must be run from the project root if using the default arguments." +
+  "\n" +
+  "\nUsage:" +
+  `\n\tdeno run -A .github/workflows/scripts/build_npm.ts -v "v0.0.0-test" [-e ${entryPointDefault}] [-o ${outDirDefault}]` +
+  "\n\tdeno run -A .github/workflows/scripts/build_npm.ts --help" +
```
hardy613: does DTN load dynamic libs? I see `-A` from their docs, but I wonder if we can tighten up permissions

.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
+import { build, emptyDir } from "https://deno.land/x/dnt@0.22.0/mod.ts";
+import { parse } from "https://deno.land/std@0.133.0/flags/mod.ts";
+
+const entryPointDefault = "./mod.ts";
+const outDirDefault = "./.github/workflows/dist";
+
+const helpText =
```
hardy613, line 7: fancy

.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  await emptyDir(outDir);
+
+  await build({
+    entryPoints: [entryPoint],
+    outDir,
+    compilerOptions: { target: "ES2021" },
+    shims: {
+      // see JS docs for overview and more options
```
hardy613: I'm confused, which js docs? I think a link might help as from the above docs this doesn't seem to have an impact (no option to add more shims) on the script, or maybe it was copied and pasted, so we can delete it?

.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
...
+      repository: {
+        type: "git",
+        url: "https://github.com/GameBridgeAI/ts_serialize.git",
+      },
+      author: "GameBridgeAI",
+      license: "MIT",
+      bugs: {
+        url: "https://github.com/GameBridgeAI/ts_serialize/issues",
+      },
+      homepage: "https://gamebridgeai.github.io/ts_serialize",
```
hardy613: RIP old docs site, I think we should linkto the readme or [github homepage](https://github.com/GameBridgeAI/ts_serialize/)

.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
...
+      keywords: [
+        "typescript",
+        "serialize",
+        "serialization",
+        "JSON",
+        "node",
+        "deno",
```
hardy613: let's remove these, pretty uncommon search terms

.github/workflows/ts_serialize_release.yml
```diff
@@ -41,9 +41,6 @@ jobs:
         run: deno run --allow-read --allow-write --allow-run .github/test_markdown.ts
         if: matrix.os == 'ubuntu-latest'
 
-      - name: Run package.json generation tests
-        run: .github/workflows/scripts/npm_release_files/create_npm_package_file_test.sh
```
hardy613, line 45: :wave:

.github/workflows/ts_serialize_release.yml
```diff
@@ -73,7 +70,7 @@ jobs:
           registry-url: https://registry.npmjs.org/
 
       - name: Build and test NPM module
-        run: .github/workflows/scripts/build.sh ${{ steps.get_github_tag.outputs.VERSION }}
+        run: deno run -A .github/workflows/scripts/build_npm.ts -v ${{ steps.get_github_tag.outputs.VERSION }}
```
hardy613: see other comment about permissions

.github/workflows/ts_serialize_tests.yml
```diff
@@ -7,9 +7,9 @@ name: tests
 
 on:
   push:
-    branches: ['**']
+    branches: ["**"]
   pull_request:
-    branches: ['**']
+    branches: ["**"]
```
hardy613, line 12: 🇲🇫

.github/workflows/ts_serialize_tests.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Deno lint
         run: deno lint
-
```
hardy613, line 34: how are these in the project, am I missing some magic lint things?

.github/workflows/ts_serialize_tests.yml
```diff
@@ -46,14 +46,11 @@ jobs:
...
       - name: Setup node
         uses: actions/setup-node@v1
         with:
           node-version: 16
           registry-url: https://registry.npmjs.org/
 
       - name: Build and test NPM module
-        run: .github/workflows/scripts/build.sh "v0.0.0-test"
+        run: deno run -A .github/workflows/scripts/build_npm.ts -v "v0.0.0-test"
```
hardy613: perms comment

.github/workflows/scripts/build_npm.ts
```diff
@@ -0,0 +1,81 @@
...
+        "ts_serialize",
+        "ts-serialize",
+      ],
+    },
+  });
+}
+
+await main();
```
hardy613: hm, this is weird to me because while `deno` does support top-level `await`s, there is nothing after the call, meaning `main()` alone would be good enough. The added `await` probably does nothing wild, but it makes me wonder if `main()` is even needed.

hardy613 (commented):
This is amazing, thanks @ChrisDufourMB. 

I left some comments, and a question about `main()`? with denos top-level await support, I still need to test it out, my dev laptop is charging.

.github/workflows/scripts/build_npm.ts
ChrisDufourMB: The awaits are fine, it was to allow early `returns`. Thinking about it some more I think what I was actually looking to do was something like `process.exit(1)`, rather than return. I will update it and remove the `main`

.github/workflows/ts_serialize_tests.yml
ChrisDufourMB: I'm not sure what you are asking here, what are you referring to?

.github/workflows/scripts/build_npm.ts
ChrisDufourMB: I wouldn't say they are uncommon, personally I am fine with keeping them. If someone searches for "node serialization" then this would be picked up.

Looking at other serialization projects I think we can actually add in `deserialize`. Having more search keywords doesn't really seem to be a huge issue.

.github/workflows/scripts/build_npm.ts
hardy925: cool, `main` can stay it's up to you, I do like the `deno.exit()` change idea though

.github/workflows/ts_serialize_tests.yml
hardy925: sorry my bad, some of the white space and quote changes

.github/workflows/scripts/build_npm.ts
hardy925: cool, let's do that then!

.github/workflows/ts_serialize_tests.yml
ChrisDufourMB: I think my editor did that automatically, I can revert those changes if you want. I don't think we have a lint spec for `yml` files

hardy925: let's keep the changes, they're fine IMHO

hardy613 (dismissed):
This is so great, :fire:/10

hardy613 (approved):

Issue #151, opened 2025-02-26
[BUG] - Nested objects cannot be serialized if they are set to null


WillForrestMindBridge:
**Describe the bug** Any nested object that is also a class cannot be serialized if it is set to null.
When set to undefined, it can serialize properly, but not with null.

![Image](https://github.com/user-attachments/assets/b41d18b9-b014-4984-a563-ba51a274b173)

**To Reproduce** Steps to reproduce the behaviour:

1. Create an class (A) with another class (B) as a field
2. Create the class
3. Set the field to null
4. Attempt to use .tsSerialize() or JSON.stringify or toJSON()
5. See error

**Expected behaviour** The primitive "null" is chosen as the serialization value.

**Desktop (please complete the following information):**

- OS: MacOS, linux
- Browser: FIrefox + chrome
- Version 2.0.3

hardy613:
This looks like a good issue, thanks for opening the PR. I left some smaller comments but overall it's great. Thanks again!

PR #152, opened 2025-02-26
151 Added handling for null class objects

WillForrestMindBridge:
**Description**

See issue 151 [here](https://github.com/GameBridgeAI/ts_serialize/issues/151)

- Allow the fromSerializable strategy to return the primitive null value instead of having a null pointer exception.
  - We expect to send null in the request.
- Undefined is protected at the level above, which makes sense, as we don't want to send it over the wire. No changes to be made in that case.

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/.github/workflows/ts_serialize_tests.yml
+++ b/.github/workflows/ts_serialize_tests.yml
@@ -40,12 +40,12 @@ jobs:
         run: deno cache test_deps.ts
 
       - name: Deno test
-        run: deno test --doc
+        run: deno test --doc -c tsconfig.json
 
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: 16
+          node-version: 20
           registry-url: https://registry.npmjs.org/
 
       - name: Build and test NPM module
--- a/polymorphic_test.ts
+++ b/polymorphic_test.ts
@@ -28,6 +28,7 @@ test({
 
       fail("polymorphicClassFromJSON did not error with no context");
     } catch (e) {
+      assert(e instanceof Error);
       assertEquals(e.message, ERROR_FAILED_TO_RESOLVE_POLYMORPHIC_CLASS);
     }
   },
@@ -210,7 +211,7 @@ test({
     class TestClass extends AbstractClass {
       @SerializeProperty("class")
       @PolymorphicSwitch(() => new TestClass(), "TestClass")
-      public [symbol]: string;
+      public [symbol]: string = "";
 
       @SerializeProperty()
       public someProperty?: string;
@@ -344,7 +345,7 @@ test({
       @SerializeProperty()
       public someProperty = "original value";
 
-      public tsTransformKey(key: string): string {
+      public override tsTransformKey(key: string): string {
         return "+" + key;
       }
     }
@@ -361,7 +362,7 @@ test({
   name: "polymorphic switch supports inherited custom tsTransformKeys",
   fn() {
     abstract class AbstractClass extends Serializable {
-      public tsTransformKey(key: string): string {
+      public override tsTransformKey(key: string): string {
         return "!" + key;
       }
     }
--- a/serializable.ts
+++ b/serializable.ts
@@ -126,8 +126,7 @@ export function toPojo(
 
   if (!serializablePropertyMap) {
     throw new Error(
-      `${ERROR_MISSING_PROPERTIES_MAP}: ${context?.constructor
-        ?.prototype}`,
+      `${ERROR_MISSING_PROPERTIES_MAP}: ${context?.constructor?.prototype}`,
     );
   }
   const record: JSONObject = {};
--- a/strategy/to_json/from_serializable.ts
+++ b/strategy/to_json/from_serializable.ts
@@ -10,6 +10,10 @@ export function fromSerializable(): ToJSONStrategy {
   ): JSONValue => {
     if (Array.isArray(value)) {
       return value.map((item) => item.tsSerialize());
+    } else if (value === null) {
+      // Objects can be null, return them without further serialization.
+      // Double cast required to return a JSONValue
+      return value as null as JSONValue;
     }
     return value.tsSerialize();
   };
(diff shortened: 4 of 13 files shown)


serializable.ts
```diff
@@ -126,8 +126,7 @@ export function toPojo(
 
   if (!serializablePropertyMap) {
     throw new Error(
-      `${ERROR_MISSING_PROPERTIES_MAP}: ${context?.constructor
-        ?.prototype}`,
+      `${ERROR_MISSING_PROPERTIES_MAP}: ${context?.constructor?.prototype}`,
```
WillForrestMindBridge, line 129: Deno format wanted this formatted. 

Wasn't my idea, but if that's what it wants, that's what it gets.

hardy925: All good.

.github/workflows/ts_serialize_tests.yml
```diff
@@ -40,12 +40,12 @@ jobs:
...
 
       - name: Deno test
-        run: deno test --doc
+        run: deno test --doc -c tsconfig.json
 
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: 16
+          node-version: 20
```
WillForrestMindBridge, line 48: If this doesn't work smoothly, I'll bump it back to 18. (and then 16 if that doesnt work nicely)

16 is ancient.

.github/workflows/ts_serialize_tests.yml
```diff
@@ -40,12 +40,12 @@ jobs:
         run: deno cache test_deps.ts
 
       - name: Deno test
-        run: deno test --doc
+        run: deno test --doc -c tsconfig.json
```
WillForrestMindBridge, line 43: when running deno tests, we need to read the tsconfig to enable experimental decorators.

Otherwise we get an angry typescript 
```
TS1240 [ERROR]: Unable to resolve signature of property decorator when called as an expression.
  Argument of type 'undefined' is not assignable to parameter of type 'Object'.
      @SerializeProperty()
```

on every decorator.

tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
-    "experimentalDecorators": true
+    "target": "ES5",
```
WillForrestMindBridge, line 3: This combination of es5 plus experimental decorators seem to do the trick locally.

A bit of a shame, I'd rather not have these at all.

WillForrestMindBridge:
Right now, there is a type error on many fields. I have not figured out a nice way to appease typescript yet. Working on it.

performing a 
`deno test -c tsconfig.json --no-check`
shows that all tests are passing.

EDIT: got it.

tsconfig.json
hardy613: @WillForrestMindBridge see [my pr](https://github.com/GameBridgeAI/ts_serialize/pull/153/files#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357) I don't think `target: ES5` is needed, we needed to tell the deno engine in vscode where the tsconfig file is located.

hardy613 (commented):
would you mind consuming my PR? then we can have this PR focus on fixing the issue raised and my PR will focus on the various engine upgrades. 

If that makes sense. 

let me know what you think

hardy613:
Also Thanks for the first contribution - please don't forget to add yourself to the `CONTRIBUTORS.md` file!

polymorphic_test.ts
```diff
@@ -28,6 +28,7 @@ test({
 
       fail("polymorphicClassFromJSON did not error with no context");
     } catch (e) {
+      assert(e instanceof Error);
```
hardy613, line 31: oh if this works I should change what I did, this is cleaner IMHO

strategy/to_json/from_serializable.ts
```diff
@@ -10,6 +10,10 @@ export function fromSerializable(): ToJSONStrategy {
   ): JSONValue => {
     if (Array.isArray(value)) {
       return value.map((item) => item.tsSerialize());
+    } else if (value === null) {
+      // Objects can be null, return them without further serialization.
+      // Double cast required to return a JSONValue
+      return value as null as JSONValue;
```
hardy613, line 16: interesting we need the double cast when the compiler can infer the first cast from the `if`

WillForrestMindBridge: Yea, 

additionally, what happens is the linter has an infinite recursion error occur when I view `JSONObject` in VSCode, unsure if it's related.

I did not find a cleaner way to do this yet.

polymorphic_test.ts
WillForrestMindBridge: This works!

WillForrestMindBridge:
@hardy613 I don't mind rebuilding this PR ontop of your other changes. The image changes etc probably shouldn't be in the fix. I was in a bit of a hurry, and this could be a smaller PR.

What happened was we had to roll back to version 1.x for stability, and while this PR works well, it has been mildly forgotten (By mildy, probably we won't switch over to version 2.x until we have some time in summer).

strategy/to_json/from_serializable.ts
hardy613: interesting error then!

.github/workflows/ts_serialize_tests.yml
hardy613: I put it up to 22 in my PR ;)

polymorphic_test.ts
hardy613: done! made the changes on my PR, thanks Will!

.github/workflows/ts_serialize_tests.yml
hardy613: see my comment on vscode settings `deno/deno.config`

WillForrestMindBridge:
Closing in favor of `protect_against_null_serialization_v2`

PR #153, opened 2025-05-08
Support deno 2.x+ 

hardy613:
**Description**

Fixes various issues with engine upgrades, mainly adds support for deno 2.x+ and keeps deno 1.x support with node 22 support

When deno 2.x is merged we should also make that the requirement for tests in place of deno 1.x tests.

- support deno 2.x+
- support node 22
- add typescript hints for overrides
- `catch(e)` where `e` is now `unknown` previously `any`
- adds `-c tsconfig.json` option to ci/cd commands
- fmt's all the codes
- updates codeQL engine to v3
- add member accessor

**Things to look at**

- [x] Test coverage
- [x] Code Style


CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v2.1.0] - 2025-05-08
```
WillForrestMindBridge, line 9: If I redo Issue #151, and point it at this branch, would I be able to slip it into the 2.1.0 release so I don't have to do a 2.1.1 release soon after?

If you want this out sooner, that's cool too.

polymorphic_test.ts
```diff
@@ -20,14 +20,14 @@ test({
...
-    class Test extends Serializable {
-    }
+    class Test extends Serializable {}
 
     try {
       polymorphicClassFromJSON(Test, {});
 
       fail("polymorphicClassFromJSON did not error with no context");
     } catch (e) {
+      assert(e instanceof Error);
```
WillForrestMindBridge, line 30: 👍

polymorphic_test.ts
```diff
@@ -210,13 +206,13 @@ test({
     class TestClass extends AbstractClass {
       @SerializeProperty("class")
       @PolymorphicSwitch(() => new TestClass(), "TestClass")
-      public [symbol]: string;
+      public [symbol]!: string;
```
WillForrestMindBridge, line 209: Is this a `definite assignment` out in the wild? [typescript doc](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions)

.github/workflows/ts_serialize_release.yml
```diff
@@ -25,17 +25,17 @@ jobs:
           deno-version: ${{ matrix.deno }}
 
       - name: Deno lint
-        run: deno lint
+        run: deno lint -c tsconfig.json
```
hardy613, line 28: thanks @WillForrestMindBridge

WillForrestMindBridge (dismissed):
LGTM

CHANGELOG.md
WillForrestMindBridge: EDIT* I can probably get this done by tomorrow, just have to switch contexts

hardy613: slipping it in to 2.1.0 works for me, thanks!

polymorphic_test.ts
hardy613: if you don't have `!` then tsc expects a value at definition or in the constructor, this class did neither (and doesn't need to set it) so we can use `!` to let tsc know it'll be set when we use it.

hardy613: FYI this is super helpful with `Maps`

```ts
const map = new Map<number, number>();
map.set(1,1);
const value = map.get(1)!; //this `!` means is guaranteed to be set
```

WillForrestMindBridge (approved):

ms1111 (approved):
Looks good!

PR #155, opened 2025-05-08
fixes #151

### LINKED ISSUE #151: [BUG] - Nested objects cannot be serialized if they are set to null

WillForrestMindBridge:
**Description**

### Issues fixed ###
See issue #151 

Issue 151
- Allow the fromSerializable strategy to return the primitive null value instead of having a null pointer exception.
  - We expect to send null in the request.
- Undefined is protected at the level above, which makes sense, as we don't want to send it over the wire. No changes to be made in that case.

**Things to look at**

- [x] Test coverage
- [x] Code Style

### DIFF
--- a/.github/CONTRIBUTORS.md
+++ b/.github/CONTRIBUTORS.md
@@ -7,3 +7,4 @@
 - [@rgrosset](https://github.com/rgrosset) :shipit:
 - [@ms1111](https://github.com/ms1111) :eyes:
 - [@khan-saim](https://github.com/khan-saim)
+- [@WillForrestMindBridge](https://github.com/WillForrestMindBridge)
--- a/strategy/to_json/from_serializable.ts
+++ b/strategy/to_json/from_serializable.ts
@@ -9,6 +9,8 @@ export function fromSerializable(): ToJSONStrategy {
     if (Array.isArray(value)) {
       return value.map((item) => item.tsSerialize());
     }
-    return value.tsSerialize();
+    // Objects can be null, return them without further serialization.
+    // Double cast required to return a JSONValue
+    return value ? value.tsSerialize() : value as null as JSONValue;
   };
 }
--- a/strategy/to_json/from_serializable_test.ts
+++ b/strategy/to_json/from_serializable_test.ts
@@ -44,6 +44,33 @@ test({
 });
 
 test({
+  name:
+    "fromSerializable - arrays of nested objects with an object set to null",
+  fn() {
+    class Test1 extends Serializable {
+      @SerializeProperty("nested_property")
+      public serializeMe = 999;
+    }
+    class Test2 extends Serializable {
+      @FromSerializable("outer_property")
+      public nested: Test1[] | null = [new Test1()];
+    }
+
+    class Test3 extends Serializable {
+      @FromSerializable("outer_outer_property")
+      public nested2: Test2[] | null = [new Test2()];
+    }
+    const testObj = new Test3();
+    testObj.nested2 = null;
+
+    assertEquals(
+      testObj.toJSON(),
+      `{"outer_outer_property":null}`,
+    );
+  },
+});
+
+test({
   name: "fromSerializable - single serializable objects",
   fn() {
     class Test1 extends Serializable {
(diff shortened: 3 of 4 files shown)


WillForrestMindBridge:
@hardy613  I was going to push this branch onto your remote fork, but github didn't let me. Nothing can ever be simple right?

I've added the 151 fix onto your changes.

hardy613:
That makes sense, the order will be to merge mine, then this should clean up to only what is needed to fix #151 and we can merge it, then @ChrisDufourMB can release

hardy925:
Did you add yourself to the contributors.md file? You should. Reporting and fixing a bug is great work. Thanks again Will.

hardy613:
This PR is a child of #153 the bulk of the specific change Will is making is in the commit [08f2aa2e](https://github.com/GameBridgeAI/ts_serialize/pull/155/commits/08f2aa2eeed11a0c5483d948f3d1d35a93c34de6)

strategy/to_json/from_serializable.ts
```diff
@@ -8,6 +8,10 @@ export function fromSerializable(): ToJSONStrategy {
   return (value: Serializable | Serializable[]): JSONValue => {
     if (Array.isArray(value)) {
       return value.map((item) => item.tsSerialize());
+    } else if (value === null) {
+      // Objects can be null, return them without further serialization.
+      // Double cast required to return a JSONValue
+      return value as null as JSONValue;
     }
     return value.tsSerialize();
```
hardy613: Hmm, I don't think we need the `else if`, it could be another `if`, this is because the above `if` `return`'s and exits anyways so an `else` is not needed. I am also thinking about something like `return value ? value.tsSerialize() : value;`

what are your thoughts? Nice comments to btw!

strategy/to_json/from_serializable_test.ts
```diff
@@ -43,6 +43,33 @@ test({
   },
 });
 
+test({
+  name:
+    "fromSerializable - arrays of nested objects with an object set to null",
+  fn() {
+    class Test1 extends Serializable {
+      @SerializeProperty("nested_property")
+      serializeMe = 999;
```
hardy613: NIT: if we happen to go back for code changes in the PR could we add `public ` to these members, only for consistency, again it's not a blocker.

.github/CONTRIBUTORS.md
```diff
@@ -7,3 +7,4 @@
 - [@rgrosset](https://github.com/rgrosset) :shipit:
 - [@ms1111](https://github.com/ms1111) :eyes:
 - [@khan-saim](https://github.com/khan-saim)
+- [@WillForrestMindBridge](https://github.com/WillForrestMindBridge)
```
hardy613, line 10: :fire:

hardy613 (commented):
:+1: 

should fix #151, thanks!

strategy/to_json/from_serializable_test.ts
WillForrestMindBridge: I have it open still, 1 second...

strategy/to_json/from_serializable.ts
WillForrestMindBridge: Sure

WillForrestMindBridge:
Squashed the PR changes, so fix 151 is still a single commit.

strategy/to_json/from_serializable.ts
WillForrestMindBridge: 😅 

I changed to

```
    return value ? value.tsSerialize() : value as null as JSONValue;
```

strategy/to_json/from_serializable.ts
```diff
@@ -9,6 +9,8 @@ export function fromSerializable(): ToJSONStrategy {
     if (Array.isArray(value)) {
       return value.map((item) => item.tsSerialize());
     }
-    return value.tsSerialize();
+    // Objects can be null, return them without further serialization.
+    // Double cast required to return a JSONValue
+    return value ? value.tsSerialize() : value as null as JSONValue;
```
hardy613, line 14: this works too, the comments are what really helps us here, so thanks again for those.

hardy613 (dismissed):
LGTM - let's get #153 in first though

hardy613 (approved):
Thanks @WillForrestMindBridge !

PR #157, opened 2025-05-09
fixes #156 DNT builds broken

### LINKED ISSUE #156: [BUG] - DNT builds not working
**Describe the bug** A clear and concise description of what the bug is.

1. run `$ ./_build_npm.ts -v v0.0.0-test -e ./mod.ts -o ./dist`
2. see error 

Image

**Expected behaviour** A clear and concise description of what you expected to
happen.

**Additional context** Add any other context about the problem here.

hardy613:
**Description**

Builds for node didn't output locally - Use JSR with a deno.json and deno.lock file. Use the new shiny standard library for JavaScript via JSR. 

fixes #156 

- 100% test coverage
- use deno `task` for tasks
- test use the `Deno.test` function for shimmed tests with DNT
- add deno.json
- add deno.lock
- remove test_deps.ts and test_deps_test.ts
- add jsr:@std/assert
- add jsr:@std/cli
- add jsr:@deno/dnt
- remove `const` from type `T` defs in polymorphic 
image
- remove `cmd` from `--allow-run` option

**Things to look at**

- pull the branch and run `$ ../_build_npm.ts -v v0.0.0-test -e ./mod.ts -o ./dist` and observe a complete build


polymorphic.ts
```diff
@@ -204,7 +204,7 @@ function resolvePolymorphicSwitch(
 /** Uses either the polymorphic resolver or the polymorphic switch resolver to determine the
  * appropriate class, then deserialize the input using Serializable#fromJSON, returning the result
  */
-export function polymorphicClassFromJSON<const T extends Serializable>(
```
WillForrestMindBridge, line 207: to be fair, const in a generic kinda feels funky, not surprised the compiler doesn't like it.

What I am suprised about is it works on github, but not locally with that command.

_build_npm.ts
```diff
@@ -94,5 +94,6 @@ try {
   await Deno.copyFile("README.md", `${flags.o}/README.md`);
   await Deno.copyFile("CHANGELOG.md", `${flags.o}/CHANGELOG.md`);
 } catch (e) {
-  printHelpText(e.message);
+  const { message } = e instanceof Error ? e : { message: `${e}` };
```
WillForrestMindBridge, line 99: 👍

WillForrestMindBridge (dismissed):

deno.json
```diff
@@ -0,0 +1,10 @@
+{
+  "imports": {
+    "@deno/dnt": "jsr:@deno/dnt@^0.42.1",
+    "@std/assert": "jsr:@std/assert@^1.0.13",
+    "@std/cli": "jsr:@std/cli@^1.0.17"
+  },
+  "compilerOptions": {
+    "experimentalDecorators": true
+  }
```
hardy613: tsconfig.json is not needed now!!

deno.json
```diff
@@ -0,0 +1,10 @@
+{
+  "imports": {
+    "@deno/dnt": "jsr:@deno/dnt@^0.42.1",
+    "@std/assert": "jsr:@std/assert@^1.0.13",
+    "@std/cli": "jsr:@std/cli@^1.0.17"
```
hardy613, line 5: jsr is superacool

`@std` is amazing to have. node supports it now too

deno.lock
hardy613, line 1: hello darkness my old friend

polymorphic.ts
hardy613: not sure it would work as well as this would enforce[ runtime generic member types ](https://xebia.com/blog/typescript-5-0-and-the-new-const-modifier-on-type-parameters/)

serialize_property.ts
```diff
@@ -54,7 +54,7 @@ interface SerializePropertyArgumentObject {
 export function SerializeProperty(
   args?: string | SerializePropertyArgument,
 ): PropertyDecorator {
-  return function (target: unknown, propertyName: string | symbol): void {
+  return (target: unknown, propertyName: string | symbol): void => {
```
hardy613, line 57: went back to legacy decorators [see our discussion](https://github.com/GameBridgeAI/ts_serialize/discussions/158)

tsconfig.json
hardy613, line 1: :wave:

test_deps_test.ts
hardy613, line 1: :wave:

test_deps.ts
hardy613, line 1: :wave:

hardy613 (commented):
@PizzaCatKing @ms1111 @WillForrestMindBridge 

I added jsr packages and fixed DNT issues, I would love for this package to also be added to the [JSR](https://jsr.io/) as well. We will also need to update the required tests to be the v2 ones now as [v1 is not supported](https://github.com/GameBridgeAI/ts_serialize/blob/develop/SECURITY.md).

deno.json
WillForrestMindBridge: Do we want experimentalDecorators on again?

What it offers changes release by release.

We could try targeting `ES7` which would be pretty stable and we wouldn't have to worry for a while.

hardy613: see the discussion https://github.com/GameBridgeAI/ts_serialize/discussions/158

WillForrestMindBridge:
Is there anything that has to be done with the v1 tests on the github side of things?

I see them on the pr, as never running.

hardy613:
@ChrisDufourMB bump plz

hardy613:
> Is there anything that has to be done with the v1 tests on the github side of things?
> 
> I see them on the pr, as never running.

yea, e need to delete them and use the v2 tests as the required, the reason is noted, but v5 of deno.lock is not supported by v1 of deno

_build_npm.ts
```diff
@@ -1,9 +1,9 @@
-#!/usr/bin/env -S deno run --allow-env --allow-read --allow-write --allow-run=npm,cmd --allow-net=deno.land
+#!/usr/bin/env -S deno run --allow-env --allow-read --allow-write --allow-run=npm --allow-net=deno.land,jsr.io
```
ChrisDufourMB: We may want to consider being even more strict with `--allow-net`. From what I see the argument allows you to specify specific jsr dependancies:
`--allow-net jsr:@std/http/file-server` (got this from the `deno compile --help` docs).

The current restrictions may allow other unknown dependancies from the same repo to do their evil deeds or whatever, though this may be mitigated by the lockfile.

This isn't mentioned in the deno docs though... further investigation may be required...

deno.lock
```diff
@@ -0,0 +1,25 @@
+{
+  "version": "5",
```
ChrisDufourMB: How are we using the lockfile?

From what I understand having a lockfile in deno only prevents changes to the provided modules, as you would expect. We can go further though and make the lockfile frozen, which would prevent modules from updating in place on install, unless explicitly called with `--frozen=false`, which will update the lockfile. This may be preferrable in a CI environemnt. I'm not super familliar with this though, so that's here may be good.

I also tried running `deno install`, which went ahead and updated the lockfile, but also added all the sub-dependances for the dependancies, which weren't there before. Specifically the dependancies for `@deno/dnt` and some others were added.

Can you please explain what you envision the process of updating the dependancies will be if, for example, a new version of `@std/cli` is released and we would like to update to it?

ChrisDufourMB (commented):
Thanks for all of this!

Looked though the tests and everything looks good, only have a few comments about the npm build script and the lockfile.

_build_npm.ts
hardy613: sounds good. 

I'll leave a more detailed comment on these files and changes, but locking down the script to known imports instead of the whole Javascript registry is a good idea :facepalm: I should have thought of that.

deno.lock
hardy613: > How are we using the lockfile?

Really we shouldn't because we describe ourselves as `zero deps`, however, we use three modules. One from `@deno` and two from `@std`. All imports are only used for testing/CI/CD.

```json
{
  "imports": {
    "@deno/dnt": "jsr:@deno/dnt@^0.42.1",
    "@std/assert": "jsr:@std/assert@^1.0.13",
    "@std/cli": "jsr:@std/cli@^1.0.17"
  }
}
```
- `@deno/dnt` is used within our `_build_npm.ts`, this module transposes deno to node
- `@std/cli` is also used in `_build_npm.ts`, `parseArgs()` for cli options
- `@std/assert` is used for `_test.ts` files

Each module reference was added with `deno add <reference>` we could also use `jsr:@std/assert` directly as an import, but this also creates the lock file.

I saw the `--frozen` option but did not build a snowman with Elsa. If we continue with a lock file we can add that as well!

From the docs we can add to `deno.json`.
```json 
{
    "lock": false
}
```
This would cause deno to skip the integrity checks when downloading the modules on a new machine, but then also allow v1 support. Thinking about the impact, I don't think in production there would be any, only for development. the deno module is zero deps and the npm module is pre-compiled. 

I should have RTFM'd a little more before going wild with the new bells and whistles.

If it makes sense with you to add the `"lock": false` option to `deno.json` I'll just do that and scrub the `deno.lock` from the github cache.

Let me know your thoughts.

hardy613:
@ChrisDufourMB :pray: thanks for the detailed review. Apologies it took an extra day or two to get back to you. The questions you are asking caused me to go back to the docs as well. These features are new-ish with deno and I was learning as I went pulling references from my experience with NPM's lock file management which is why I committed it. 

I left responses, I think we have a way forward, but I'll hold for your thoughts. I like the idea of dropping the lock file all together on this project as production code is zero dependencies. 

Thanks again for taking the time to leave an awesome review.

Cheers :taco: :taco: :taco: :clock630:

_build_npm.ts
hardy613: I tried this and it did not work. 

image


further reading of the docs, `jsr.io` is enabled by default. seems like [a trusted list exists](https://docs.deno.com/runtime/fundamentals/security/): 

    https://deno.land/
    https://jsr.io/
    https://esm.sh/
    https://raw.githubusercontent.com
    https://gist.githubusercontent.com

.github/workflows/ts_serialize_tests.yml
```diff
@@ -30,17 +30,23 @@ jobs:
...
+        run: deno task test:coverage
+
+      - name: Generate coverage report
+        run: deno task test:coverage:report
+
+      - name: Coverage report
+        uses: VeryGoodOpenSource/very_good_coverage@v2
+        with:
+          path: "./coverage/lcov.info"
+          min_coverage: 100
```
hardy613, line 71: :100:

.github/workflows/ts_serialize_tests.yml
```diff
@@ -30,17 +30,23 @@ jobs:
...
-        run: deno fmt --check -c tsconfig.json
+        run: deno task fmt --check
         if: matrix.os != 'windows-latest'
 
-      - name: Cache dependencies
-        run: deno cache test_deps.ts
-
       - name: Deno test
-        run: deno test --doc -c tsconfig.json
+        run: deno task test:coverage
```
hardy613, line 62: `task` uses deno.json auto discovery

strategy/utils.ts
```diff
@@ -7,10 +7,15 @@ export type NewSerializable<T> = T & (new () => Serializable);
...
-  } else if (isFunctionSerializable(type)) {
-    return type();
+  } else if (!isNewableType && isFunctionSerializable(type)) {
+    const instance = type();
+    if (instance instanceof Serializable) {
+      return instance;
+    }
   }
+
   throw new Error(ERROR_GET_NEW_SERIALIZABLE_SERIALIZABLE_NOT_RETURNED);
```
hardy613, line 20: actually fixed a bug where the error was never reached, thanks test coverage!

strategy/utils_test.ts
```diff
@@ -39,3 +39,23 @@ test({
...
+      fail("Expected getNewSerializable error to be thrown");
+    } catch (e) {
+      assert(e instanceof Error);
+      assertEquals(
+        e.message,
+        ERROR_GET_NEW_SERIALIZABLE_SERIALIZABLE_NOT_RETURNED,
+      );
+    }
+  },
+});
```
hardy613, line 61: 100% coverage

hardy613 (commented):
@ChrisDufourMB i went ahead and made the changes

hardy613:
Okie doke, I think I got it all now, v1 tests, v2 tests++, hopefully when v3 is out it won't affect ci as much.

strategy/to_json/from_serializable.ts
```diff
@@ -5,12 +5,10 @@ import { ToJSONStrategy } from "../compose_strategy.ts";
...
 /** serialize data using `tsSerialize` on a subclass Serializable type */
 export function fromSerializable(): ToJSONStrategy {
-  return (value: Serializable | Serializable[]): JSONValue => {
+  return (value: Serializable[] | null): JSONValue => {
     if (Array.isArray(value)) {
       return value.map((item) => item.tsSerialize());
     }
-    // Objects can be null, return them without further serialization.
-    // Double cast required to return a JSONValue
-    return value ? value.tsSerialize() : value as null as JSONValue;
```
hardy613, line 14: turns out this is never reached as `toPojo(serializable)` will be recursive on `Serializable`'s

SECURITY.md
```diff
@@ -4,11 +4,20 @@ Below is a table of supported versions.
...
 | ------- | ------------------ |
 | 2.x     | :white_check_mark: |
 | 1.x     | :x:                |
 
+### Deno
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 2.x     | :white_check_mark: |
+| 1.x     | :white_check_mark: |
```
hardy613, line 19: feels good

_build_npm.ts
```diff
@@ -48,13 +47,16 @@ if (flags.h) {
...
+
 try {
   await emptyDir(flags.o);
 
   await build({
+    importMap: "./deno.json",
     entryPoints: [flags.e],
     outDir: flags.o,
-    compilerOptions: { target: "ES2021" },
+    compilerOptions: { lib: ["ESNext"], experimentalDecorators: true },
```
WillForrestMindBridge, line 59: I always get a little unnerved when the target is unnumbered like "ESNext".

When does ESNext stop 😂?

strategy/to_json/from_serializable.ts
WillForrestMindBridge: ⚠️ 

Is an array of nulls possible now? Should we write a test to check that case?

Eg.

```
value = [null,null,null,null]
```

I suppose that was always possible, just noticing now that you've cleaned up a bit.

_build_npm.ts
hardy613: image


yea without it the npm build fail missing types on Sets and ErrorOptions

strategy/to_json/from_serializable.ts
hardy613: :100: nice catch, adding the test found a bug, it's fixed now and we will support this request next release, thanks @WillForrestMindBridge

strategy/to_json/from_serializable_test.ts
```diff
@@ -122,3 +122,28 @@ test({
     );
   },
 });
+
+Deno.test({
```
WillForrestMindBridge, line 126: Nice 👍

WillForrestMindBridge (approved):
The othes may be a while before they get to review. They are extremely busy until end of month.

I look forward to trying out tsSerialize v2 in a bigger project.

hardy613:
@ChrisDufourMB bump please.

ChrisDufourMB (approved):

PR #165, opened 2025-07-07
JSR readiness

hardy613:
**Description**

TBA - This PR is part of a larger discussion about how and where we publish this repo. 

https://jsr.io/


hardy925:
@ms1111 I see you're in the repo lately. Could I get your thoughts here on JSR? IF we keep distributing to NPM could we also add JSR - @ChrisDufourMB followed up about it and I hear MB devs didn't want to move away from NPM - which I get, but if we keep NPM publishing there is no need to say no to also publishing on JSR (a superset of NPM)

Thanks in advance.

ms1111:
I took a read through the JSR docs today and I liked that they started out with OAuth based publishing (trusted publishing). I also like that you can register a scope and delegate it to a few GitHub users.

Let me ask some questions on my end to see if there would be any problems with publishing to JSR. I might need a week or two.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v4
+      # using npx to avoid needing to install deno
+
+      # --dry-run is used to avoid publishing the
+      # module until we can figure out how to handle
+      # the jsr.io publishing with the @gamebridge account admin.
+
+      # When we decide to publish to JSR we'll need to rename the
+      # module to @gamebridge/serialize or something similar
+      # because `_` is not allowed in module names.
```
ms1111, line 24: Oh, that's interesting.

.github/workflows/ci.yml
```diff
@@ -70,12 +44,6 @@ jobs:
...
-        uses: actions/setup-node@v1
-        with:
-          node-version: 22
-          registry-url: https://registry.npmjs.org/
-
-      - name: Build and test NPM module
-        run: deno task build
+      - name: Publish module to jsr.io
+        run: deno publish --dry-run
         if: matrix.os == 'ubuntu-latest'
```
ms1111, line 49: Could we publish to both jsr and npm? Is that a typical thing to do?

.github/workflows/publish.yml
hardy613: yeaa! Man it was a gut punch to hear that

.github/workflows/ci.yml
hardy613: It is pretty normal to publish to both. NPM adoption is important, and JSR is described as a superset of NPM.

hardy613: Although in my eagerness, I removed NPM publish, I can easily put it back

hardy925:
Hey @ms1111, I hope things are well.

Any further news on JSR adoption?

PR #166, opened 2025-09-26
Configure GitHub/npmjs trusted publishing

ms1111:
**Description**

npm added Trusted Publishing a few months back. It uses OpenID Connect to manage an automatically rotated token between GitHub and npmjs.

Given recent attacks on the npm ecosystem, I'd like to adopt this. I've set up the link on the npmjs side following the docs at https://docs.npmjs.com/trusted-publishers

- When publishing to NPM, use the Trusted Publishing token instead of the static NPM token.

**Things to look at**

- [ ] Test coverage
- [ ] Code Style

### DIFF
--- a/.github/workflows/ts_serialize_release.yml
+++ b/.github/workflows/ts_serialize_release.yml
@@ -6,6 +6,11 @@ on:
   push:
     tags: ["v*.*.*"]
 
+# OIDC setup for trusted publishing
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
@@ -76,9 +81,10 @@ jobs:
           draft: false
           prerelease: false
 
+      - name: Update npm to 11.5.1 or later for trusted publishing
+        run: npm install -g npm@latest
+
       - name: Deploy to NPM
         run: |
           cd dist
           npm publish --access public
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}


hardy925 (approved):
Great! thanks for the contribution @ms1111 

> Given recent attacks on the npm ecosystem

I am familiar with these. Unfortunately this will not fix the type of supply chain attack both of these CVEs start from (both attacks from the same team BTW)

The first attack was done via NX scripts and they got 5 whole cents! https://www.securityalliance.org/news/2025-09-npm-supply-chain

the second was from tinycolor and was much worse because it would infect your CI and spread to other packages that way: https://www.ox.security/blog/npm-2-0-hack-40-npm-packages-hit-in-major-supply-chain-attack/

Either way in both attacks a newly update patch was made in a way to be leverage npms `~` and `^` and how developers manage their packages. Remember my time working together I proclaimed we needed to `pin` our packages to the exact version we want to use. This is the exact reason why. It is more work and requires regular maintenance to stay up-to-day, but it's the most secure way as patches are not silently consumed.

Our build uses no imports (zero deps baaaaabyyyyy!) However in tests we make uses of JavaScripts standard library hosted on JSR

Then in the script to make the NPM build we use a deno package DNT - all our devDeps in the `deno.json` file:
```json
{
  "imports": {
    "@deno/dnt": "jsr:@deno/dnt@^0.42.1",
    "@std/assert": "jsr:@std/assert@^1.0.13",
    "@std/cli": "jsr:@std/cli@^1.0.17"
  },
  //...
}
```

These are what this repo in particular needs to watch out for, but given it is three imports and two are from the `JS @std` library, it's reasonable to manage. 

Again thanks @ms1111, this is a welcome change.

hardy925:
@ms1111 I am realizing that we didn't `pin` package in this repo so tech we are still open to a supply chain attack. If you're hungry for more code, updating the packages to be pinned would be awesome. 

A changelog.md entry for this would also be welcome.

ms1111:
Hey, thanks @hardy925 ! I can take a look at pinning those three deps in deno.json.

ms1111:
@hardy925 Let me actually see if I can do that in a separate pull request, just in case it gets gnarly - I'm going to set `lock` to true, and I don't really know what I'm doing with deno yet.

hardy925:
> @hardy925 Let me actually see if I can do that in a separate pull request, just in case it gets gnarly - I'm going to set `lock` to true, and I don't really know what I'm doing with deno yet.

yea sounds good to me! thanks @ms1111

hardy613 (approved):

PR #167, opened 2025-09-29
Pin packages and add lockfile for reproducible builds

ms1111:
**Description**

To ensure reproducibility and minimize susceptibility to supply chain attacks, use a lock file for dependencies. As suggested in https://github.com/GameBridgeAI/ts_serialize/pull/166#issuecomment-3339625062

Using lockfile for deno v2 only.

For v1, it'd be a different lockfile format, and given it's only a few, pinned, standard packages, the risk of using unpinned transitive dependencies is low, and I figure the v1 build will go away at some point.

**Things to look at**

- [ ] Test coverage
- [ ] Code Style

### DIFF
--- a/.github/DEVELOPMENT.md
+++ b/.github/DEVELOPMENT.md
@@ -25,13 +25,13 @@ This will add the deno types to the project.
 ## Running the tests
 
 ```bash
-$ deno test
+$ deno task test
 ```
 
 You can add a file to test only that one file.
 
 ```bash
-$ deno test mod_test.ts
+$ deno test --frozen mod_test.ts
 ```
 
 ### Coding style and tests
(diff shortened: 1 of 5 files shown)


hardy925 (approved):
LGTM

Thanks for also updating the packages. 

In terms of the lock file I think @ChrisDufourMB did some work / research into it before so it might be useful to get his input as well. 

image

hardy613 (approved):
c.c @ChrisDufourMB @PizzaCatKing

hardy613:
> and I figure the v1 build will go away at some point.

yea, we can discuss that deprecation, although I hear v1 is still in use - so we'll need ot be clear about it somehow

hardy925 (approved):

